PR #5, opened 2020-02-13
Use isomorphic useLayoutEffect

alemesa:

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] Code style update
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## Problem Description

When we do initial animations we usually use useLayoutEffect to wait for the DOM to render since useEffect sometimes makes the website flash for a very small amount of time (like 50ms) which is noticeable on the first hard refresh. 

Using useLayoutEffect throws a warning with nextjs because it has no use in the server

## Solution Description

Swap useLayoutEffect that you want to wait for the DOM to render with useIsomorphicLayoutEffect

Someone smarter than me explained it here
https://medium.com/@alexandereardon/uselayouteffect-and-ssr-192986cdcf7a

- [x] N/A

**Aditional comments:**

You should still use useEffect as normal

We could also just create our own hook since this is essentially just this:
```js
const useIsomorphicLayoutEffect =
  typeof window !== 'undefined' ? useLayoutEffect : useEffect
```

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -1,6 +1,7 @@
-import React, { useRef, useCallback, useEffect } from 'react';
+import React, { useRef, useCallback } from 'react';
 import Head from 'next/head';
 import { useDispatch } from 'react-redux';
+import { useIsomorphicLayoutEffect } from 'react-use';
 import { gsap } from 'gsap';
 
 import styles from './index.module.scss';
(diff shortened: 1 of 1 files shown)


neo:
:spinthink:

alemesa:
@neo did some research and this PR is still valid, basically nextjs adds a display: none to the body until the js loads and that's why the flashing doesn't happen 🙏 🔍 with LayoutEffect

jmckinnell:
> @neo did some research and this PR is still valid, basically nextjs adds a display: none to the body until the js loads and that's why the flashing doesn't happen 🙏 🔍 with LayoutEffect

Amazing!

PR #33, opened 2020-05-01
Feature: TypeScript

neo:
Benefit I personally see:
- Next.js provides official, built-in support for TypeScript
  - Most of the new files (e.g. `tsconfig.json` and `next-env.d.ts`) are generated by next
  - It supports both TS traspiling and type-checking in webpack
  - It also provides typings for all the next interface
- It could be progressively opted-in
  - All the `.js` files would still work as non-typed files, the ones with js-doc would also be interpreted and type checked
  - Only opt-in when naming a file with a `.ts` or `.tsx` extension
    - See the first commit for a minimum working version with just three TS files.
- It works with ESLint supported by `@typescript-eslint`


neo:
todos:
- [ ] make storybook work
- [ ] adjust the page/component template + script

DonghyukJacobJang:
closing this PR. #132

PR #34, opened 2020-05-11
New Linter ls-lint

grez96:
feature: ls lint integration
  - new package ls-lint
 - executes as part of the 'npm run linters'
 - created basic rules to support the generator's existing file naming conventions

### DIFF
--- a/.ls-lint.yml
+++ b/.ls-lint.yml
@@ -1,0 +1,16 @@
+ls:
+  .js: kebab-case | pointcase
+  .config.js: kebab-case
+  src:
+    components:
+      .js: PascalCase
+      .stories.js: PascalCase
+    pages:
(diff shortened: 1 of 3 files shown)


.ls-lint.yml
```diff
@@ -0,0 +1,16 @@
...
+      .stories.js: PascalCase
+    pages:
+      .js: kebab-case | regex:^_[a-z-.]* #regex for kebab-case with preceeding underscore
+  scripts:
+    templates:
+      .js: PascalCase
+      .stories.js: PascalCase
+
+ignore:
+  - node_modules
```
DonghyukJacobJang, line 16:
```suggestion
  - node_modules
  - .next
  - out
```
If we can add two more ignore options, it would be amazing! 👍

PR #35, opened 2020-05-11
example: add https protocol example

DonghyukJacobJang:


next.config.js
```diff
@@ -69,7 +69,7 @@ const nextJSConfig = {
     DNS_PREFETCH: process.env.DNS_PREFETCH
   },
   exportTrailingSlash: true,
-  // compress: true, // NOTE: enable this when doing SSR
```
DonghyukJacobJang, line 72: `compress` option must be disabled when running a custom server. 
Reference: https://github.com/expressjs/compression/issues/78

PR #36, opened 2020-05-11
feature: nextjs 9.4.0 and react refresh

### LINKED ISSUE #31: [Feature] Add React Refresh
https://github.com/zeit/next.js/pull/12008

DonghyukJacobJang:
- Upgrade nextjs version to 9.4.0
- Enable react refresh
- Remove the custom CSS rules that cause the break. reference: https://github.com/zeit/next.js/issues/12304
- NextJS 9.4.0 now supports `.env`. Adding the env key in next.config.js is no longer required to expose these variables.
- Remove service worker since it is hard to utilize it as expected.
- Remove `_error.js` that is deprecated.

(This closes #31.)


neo:
I think it's worth mentioning that a side effect of this change would utilize the native scss modules support that nextjs has:
- better code splitting: https://nextjs.org/blog/next-9-3#built-in-sass-css-module-support-for-component-level-styles
- add more certainty to the generated classname output

next.config.js
```diff
@@ -61,19 +40,13 @@ const optimizedImagesConfig = {
...
-    CUSTOM_ENV: process.env.CUSTOM_ENV,
-    DNS_PREFETCH: process.env.DNS_PREFETCH
-  },
   exportTrailingSlash: true,
   // compress: true, // NOTE: enable this when doing SSR
   devIndicators: {
     autoPrerender: false
   },
   experimental: {
+    reactRefresh: true,
```
neo, line 49: 9.4 has it enabled by default I think?

PR #45, opened 2020-07-25
[FIX] Environment Variables on Browser

grez96:

- [*] Bugfix (non-breaking change which fixes an issue)
- [ ] Code style update
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [*] No

- [*] I lightly tested it in one browser

## Problem Description

Environment variables are not working on the browser  when using the NEXT_PUBLIC prefix.
Screen Shot 2020-07-25 at 9 05 45 AM
Screen Shot 2020-07-25 at 9 06 42 AM

## Solution Description

https://github.com/vercel/next.js/discussions/12815

I noticed that dotenv is being used in:
```
next.config.js
nextjs.sitemap-generator.js
robots-txt-generator.js
```
But productions builds still work even after uninstalling the package. Might be worth double checking though. 👍

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,6 @@
     "babel-preset-react-app": "^9.1.1",
     "chalk": "^3.0.0",
     "cross-env": "^7.0.0",
-    "dotenv": "8.2.0",
     "eslint": "6.8.0",
     "eslint-config-postcss": "3.0.7",
     "eslint-config-prettier": "6.10.0",


neo:
I just hit this problem today 😂  trying to figure out a solution. But we'd want to opt out and do our own dotenv because we have more than two environments.

grez96:
Ahhh good catch. Just noticed that nextjs only supports
```
.env.local
.env.development
.env.production
.env.test
```
👍

neo:
apparently installing `dotenv` won't disable the default behavior anymore 🙄  according to this: https://github.com/vercel/next.js/discussions/14609#discussioncomment-30965

Issue #40, opened 2020-05-22
[Bug] withRedux doesn't work with getStaticProps


neo:
the `withRedux` wrapper which uses `getInitialProps` under the hood, isn't compatible with using `getStaticProps` because only either one or the other could be present, but not both.

neo:
maybe this could be a potential solution?
https://github.com/kirill-konshin/next-redux-wrapper#getstaticprops

neo:
I think they also updated their example: https://github.com/vercel/next.js/tree/canary/examples/with-redux

neo:
now that I have actually used this, I think for ssg we should use provide the store with the initial state in `_app.js`? because most of these complicated methods are trying to solve problems for ssr, which we don't use often.

namniak:
https://github.com/Jam3/nyg-nextjs/pull/62

PR #46, opened 2020-07-26
Nextjs with redux

DonghyukJacobJang:
#40


namniak (commented):
@DonghyukJacobJang if you initially land on a page other than landing e.g. `/about` then go to landing it will never update `loaded` state even if you go back and forth

image

PR #49, opened 2020-08-12
Feature: Access Sass Variables within Javascript

artemleznikov:
Currently when we want to access a Sass variable within JavaScript, we add it to a CSS variable and retrieve using `getPropertyValue('--var')`. Now that we're SSR'ing, we are only able to access these variables on the browser side. Due to this limitation, this feature adds the ability to import sass variables directly in our JavaScript files that works on server side.

Caveats: 
1) Requires overwriting a cssLoader option. (https://github.com/vercel/next.js/issues/11629)
2) Scss export file is required to be a `.module.scss` otherwise NextJS will throw error (https://github.com/vercel/next.js/blob/master/errors/css-global.md).

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -47,6 +47,15 @@ const nextJSConfig = {
     modern: true
   },
   webpack: function (config, options) {
+    const moduleSassRule = config.module.rules[1].oneOf.find(
+      (rule) => rule.test.toString() === /\.module\.(scss|sass)$/.toString()
+    );
+
+    if (moduleSassRule) {
+      const cssLoader = moduleSassRule.use.find(({ loader }) => loader.includes('css-loader'));
+      if (cssLoader) cssLoader.options.modules.mode = 'local';
+    }
+
     return config;
   }
 };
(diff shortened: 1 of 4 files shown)


DonghyukJacobJang:
This is nice. I have one question. Can Sass Variables be dynamic? You know that CSS Variables can be retrieved and updated in JS side when needed.

artemleznikov:
I dont think so. In that scenario, css vars would be more ideal, or toggling a body class with JS and targeting `body:global(.class) & {}` within components to update styles dynamically.

This is mostly for server side rendered html that is dependent on variables from sass. One example, in Toyota, we used sass variables from our grid system (column width, gutters, max width, breakpoints, etc.) to generate `` sizes for srcset.

namniak:
@artemleznikov @DonghyukJacobJang agreed that css vars has its place in certain scenarios but for static things like layout/breakpoints this new approach is perfect. 
Very helpful framework extension.

fpichlou:
@artemleznikov thank you so much! My problem has been fixed by using naming convention rule, the second caveat :)

PR #52, opened 2020-08-18
Feature: Add Footer component and update HTML semantics

artemleznikov:
- Create generic Footer component which holds list of all routes
- Move Nav and RotateScreen to Layout, have all top level components on the same level (Nav, Main, Footer)
- Use `<main>` tag for pages rather than section, add appropriate styles
- Add mobile/desktop height to Nav using vars
- Add simple routes file for page routes to be used within Nav and Footer

### DIFF
--- a/src/components/Layout/Layout.js
+++ b/src/components/Layout/Layout.js
@@ -1,16 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import checkProps from '@jam3/react-check-extra-props';
-import classnames from 'classnames';
 
-import styles from './Layout.module.scss';
-
-// import useLayout from '../../utils/hooks/use-layout';
+import Nav from '../Nav/Nav';
+import Footer from '../Footer/Footer';
+import RotateScreen from '../RotateScreen/RotateScreen';
 
 function Layout({ children }) {
-  // const layout = useLayout().layout;
+  return (
+    <>
+      <Nav />
 
-  return <main className={classnames(styles.Layout)}>{children}</main>;
+      {children}
+
+      <Footer />
+
+      <RotateScreen />
+    </>
+  );
 }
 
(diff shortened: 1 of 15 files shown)


namniak:
@artemleznikov @DonghyukJacobJang @iranreyes 
we need to think about the future of `jam3/react-ui` in terms of integration with the new generator.

ideally, all basic components like `Nav`, `Footer`, `RotateScreen` etc would live in one place and outside the generator. 
Currently, both Nav and Footer in react ui library accept type of a link component and default to `BaseLink` (built on top of `react-router-dom`) so we'd need to make it a bit more flexible so that Next.js syntax work with it too.

src/components/Layout/Layout.js
```diff
@@ -1,16 +1,23 @@
...
+  return (
+    <>
+      <Nav />
+
+      {children}
+
+      <Footer />
 
-  return <main className={classnames(styles.Layout)}>{children}</main>;
+      <RotateScreen />
```
namniak, line 18: we can add check here if it's a touch device

src/components/Layout/Layout.js
```diff
@@ -1,16 +1,23 @@
...
 import PropTypes from 'prop-types';
 import checkProps from '@jam3/react-check-extra-props';
-import classnames from 'classnames';
 
-import styles from './Layout.module.scss';
-
-// import useLayout from '../../utils/hooks/use-layout';
+import Nav from '../Nav/Nav';
+import Footer from '../Footer/Footer';
+import RotateScreen from '../RotateScreen/RotateScreen';
```
namniak, line 7: we should use dynamic import here with no SSR
```
const RotateScreen = dynamic(
  () => import('../RotateScreen/RotateScreen'),
  { ssr: false }
)
```

PR #55, opened 2020-08-27
Adding focus visible + cleanup

namniak:
- add `focus-visible` for tab highlight
- remove `isBrowser` in favor of using `typeof window !== 'undefined'` in place to make sure code wrapped in this condition doesn't get into the SS bundle. Webpack won't strip code portions using pre-evaluated values outside the file and will include it by default. similar to `process.env.NODE_ENV !== 'production'`, `typeof window !== 'undefined'` needs to be used in place
- general cleanup


neo (approved):
LGTM! 👍

PR #90, opened 2021-03-16
feature: updating lock body scroll util file

jxrin:

- [ ] Code style update
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## Problem Description

Update lock body scroll util file using a css variable.

## Solution Description

Updated and cleaned up the lock body scroll util file to instead lock and unlock using a global css variable.

- [x] N/A

**Aditional comments:**

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -8,7 +8,6 @@ import Head from '../components/Head/Head';
 
 import { withRedux } from '../redux/withRedux';
 import { setLandingLoaded } from '../redux/modules/app';
-import { lockBodyScroll, unlockBodyScroll } from '../utils/lock-body-scroll';
 
 function Landing() {
   const containerRef = useRef();
@@ -62,8 +61,6 @@ function Landing() {
           </li>
         </ul>
       </section>
-      <button onClick={lockBodyScroll}>Lock!</button>
-      <button onClick={unlockBodyScroll}>Unlock</button>
     </main>
   );
 }
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -8,6 +8,7 @@ import Head from '../components/Head/Head';
 
 import { withRedux } from '../redux/withRedux';
 import { setLandingLoaded } from '../redux/modules/app';
+import { lockBodyScroll, unlockBodyScroll } from '../utils/lock-body-scroll';
 
 function Landing() {
   const containerRef = useRef();
@@ -61,6 +62,8 @@ function Landing() {
           </li>
         </ul>
       </section>
+      <button onClick={lockBodyScroll}>Lock!</button>
+      <button onClick={unlockBodyScroll}>Unlock</button>
     </main>
   );
 }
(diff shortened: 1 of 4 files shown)


DonghyukJacobJang (changes requested):
This util is not only handling scroll locking/unlocking but also restore scroll position. Please re-consider its functionality and update it accordingly.

jxrin:
sorry about that! I'll update it accordingly to also store the lock position, thanks for noting it out :) @DonghyukJacobJang

jxrin:
Added all the required fixes :) please let me know what you think! @DonghyukJacobJang

DonghyukJacobJang (commented):
@jxrin Thanks for your fix. The idea was great!. However, I am afraid of approving this PR for now. There is an issue when locking and unlocking body scroll, it causes a page to be jumped x-axis due to appearing/disappearing native scroll bar on the right. 

Would you mind having a look at it?

jxrin:
@DonghyukJacobJang yeah for sure! Could you give me a bit more information for what device and browser you noticed this behaviour on, and what your procedure is for testing? I can start debugging it then 😄

DonghyukJacobJang:
@jxrin Please find an attached video as an issue reference.
I tested this on macOS 11 on Chrome 88.

https://user-images.githubusercontent.com/16929537/112197308-529fb480-8be2-11eb-9c9f-9c6cbd3c2822.mov

jxrin:
Thanks for the video @DonghyukJacobJang :) I have some time today and tomorrow so I'll look into this and see what I can change

KeshavChawla:
Hey @DonghyukJacobJang! I had some free time so I've been working on the generator. CC:: @jxrin 

I've been trying to reproduce the issue you mentioned in the video but I am unable to do so in the default settings (see video) but am able to by forcing scroll bars to always appear. By any chance do you have always on scrollbars enabled (macOS System Preferences > General > Show Scroll Bars)?

I am able to reduce the jumping by hiding the scroll bar all together:

```
::-webkit-scrollbar {
  display: none;
}
```

Is this what you meant or would you rather have the scrollbars force visibility when locked? Thanks!


https://user-images.githubusercontent.com/21203253/113174554-7da88a80-9218-11eb-9f66-1702a1800c6a.mov

DonghyukJacobJang:
@KeshavChawla thanks for debugging the issue. I am afraid of saying No regarding the scrollbar options. I am using the default which is "Automatic ...". And for the A11y purpose, setting the web scrollbar to be `display none` is not recommended.

namniak:
I think we might need to cancel this PR. as the existing solution proofed to be well working as expected in new and old browsers with scroll + lock + page restoration functionality @DonghyukJacobJang 
I don't think there's any better solution out there tbh

alemesa:
I'll close this @DonghyukJacobJang spent a bit of time this morning looking at solutions and they are not better than ours, like Vadim said this has been tested cross browsers and performs well

PR #91, opened 2021-03-18
fix: track window undefined and default gtm code snippet

DonghyukJacobJang:

- [ ] Code style update
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] N/A

**Aditional comments:**


src/utils/detect.js
```diff
@@ -5,7 +5,9 @@ const detect = {
   bots: {}
 };
 
-if (typeof window !== 'undefined') {
+export const isBrowser = typeof window !== 'undefined';
```
namniak, line 8: I would avoid exposing this via export because it will encourage people to use that instead of checking `typeof window !== 'undefined';` in place. this creates differences with webpack export and server bundle.

Webpack is doing things file-by-file that means if variable is evaluated outside the file, it will be treated as non-falsey (because it was declared globally) but webpack doesn't know the value so it will include the chunk of code in the server bundle even though it won't execute it

src/utils/track.js
```diff
@@ -1,4 +1,7 @@
-window.dataLayer = window.dataLayer || [];
+import React from 'react';
+import { isBrowser } from './detect';
+
+if (isBrowser) window.dataLayer = window.dataLayer || [];
```
namniak, line 4: see the above message. it doesn't make a lot of difference here but people will get used to using it assuming that
```
if (isBrowser) {
  // some FE code
  // some FE require
}
```
wouldn't be included in the server bundle but it would unlike using 
```
if (typeof window !== 'undefined') {
  // some FE code
  // some FE require
}
```

namniak (changes requested):

DonghyukJacobJang:
@namniak Thanks for your feedback! it has been applied!

src/components/Head/Head.js
```diff
@@ -48,6 +49,8 @@ function Head({ title, description, keywords }) {
       {/* Other recommends */}
       <link rel="canonical" href={ogUrl} />
       {process.env.NEXT_PUBLIC_DNS_PREFETCH && <link rel="dns-prefetch" href={process.env.NEXT_PUBLIC_DNS_PREFETCH} />}
+
+      {GtmScript()}
```
namniak, line 53: @DonghyukJacobJang is there a preference in adding this here vs in the `_document`?
is it the order of scripts being loaded?
also I always add it to the document and not 100% sure if having that in the component that is used in every page will cause extra initialization on route change

DonghyukJacobJang: yeah. it was because of the order of meta tags. fortunately, it will not re-init with our current setup. I can run more test.

namniak: then lgtm!

namniak (approved):

PR #92, opened 2021-03-19
feature: redux toolkit implementation

### LINKED ISSUE #73: [Feature] Replace Redux with Redux Toolkit
Easier to set up and almost half number of lines to use redux.

alemesa:
Start migrating to [redux-toolkit](https://redux-toolkit.js.org/)

Needs to be tested with SSR:

- Find a solution to SSR that looks similar to what @namniak did in the generator previously

It will resolved #73


package.json
```diff
@@ -37,6 +37,7 @@
   "dependencies": {
     "@jam3/detect": "1.0.2",
     "@jam3/react-hooks": "1.0.0",
+    "@reduxjs/toolkit": "^1.5.0",
```
neo, line 40: would be best to also remove redux as it should not be needed directly

src/pages/_app.js
```diff
@@ -8,6 +9,8 @@ import '../styles/global.scss';
 
 import Layout from '../components/Layout/Layout';
 
+import { store } from '../redux/index';
```
neo, line 12:
```suggestion
import { store } from '../redux';
```

src/redux/index.js
```diff
@@ -1,13 +1,21 @@
...
+  }
 });
 
-export const initializeStore = (preloadedState) => {
-  return createStore(reducers, preloadedState, composeWithDevTools(applyMiddleware()));
-};
+export const { setLandingLoaded } = appSlice.actions;
+
+export const store = configureStore({
+  // middleware: getDefaultMiddleware({ immutableCheck: false, serializableCheck: false }), // disabled only because the data in the state is too big
```
neo, line 18: can be removed 
```suggestion
```

src/redux/index.js
```diff
@@ -1,13 +1,21 @@
...
+  reducer: appSlice.reducer,
+  devTools: process.env.NODE_ENV !== 'production'
```
neo, line 20: this is by default like this – so if you want this, we could remove; however, like the sourcemap if you want disabled only on production production, you could consider using that instead.

DonghyukJacobJang (approved):

PR #95, opened 2021-03-22
Feature: Image Component (sizes + optimizations + webp)

artemleznikov:
- Image component utilizing `next-optimized-images` to resize, optimize images and convert them to webp.
- generates `srcSet` for image widths starting at 320 and incrementing by 320 all the way to 3840 (4k resolution)
- `sizes` object prop allows you to specify the width of image for each breakpoint to serve best image (supports px and vw values)
- has webp support by default and will fallback with a check if webp not supported 
- `OPTIMIZE_IMAGES` env variable to specify if images should be optimized (`false` on `development` to avoid long compile times)
- Images go into `src/assets/images` which Image component will retrieve using file name from `imageObj` prop


src/components/Image/Image.js
```diff
@@ -0,0 +1,82 @@
...
+const Image = forwardRef(({ className, imageObj, loadingType, sizes }, ref) => {
+  const { file, alt } = imageObj;
+  const isWebpSupported = useSelector((state) => state.app.isWebpSupported);
+
+  const { src, srcSet } = useMemo(() => {
+    const extension = file.split('.').pop();
+
+    // optimizes using regular next-optimized-images plugins
+    const src = require(`../../assets/images/${file}`);
+
```
DonghyukJacobJang, line 26: we can think of adding an error checker; the file does not exist.

alemesa: tbh this break the application and throws and error so I don't think we need an error checker here @DonghyukJacobJang

alemesa (approved):

alemesa:
Rebased and updated this PR with the redux changes @DonghyukJacobJang

DonghyukJacobJang (approved):

PR #96, opened 2021-03-30
feature: add skip to content links

KeshavChawla:

- [ ] Code style update
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

**Adviser warns of an error with netmask (high).**

Advisory code is 1658: https://www.npmjs.com/advisories/1658**

## Problem Description

Per WCAG 2.1 AA compliance having [bypass blocks](https://www.w3.org/WAI/WCAG21/quickref/#bypass-blocks) on each page is necessary.

## Solution Description

 One notably method of solving this is by:

- [G1: Adding a link at the top of each page that goes directly to the main content area](https://www.w3.org/WAI/WCAG21/Techniques/general/G1.html)

An alternate solution for Jam3 projects would include adding a subnav which satisfies criteria:

- [G124: Adding links at the top of the page to each area of the content](https://www.w3.org/WAI/WCAG21/Techniques/general/G124.html)

---- 

This PR creates a tab-able, screen readable skip to content button which the user can click to skip the navbar on each page. It does so by using an anchor link at the bottom of the nav page to change focus.

**Notable findings:**
- This link **must** be the first targetable focus on each page
- While screen readers should be able to pick up content with `opacity: 0;` [ChromeVox seems to ignore elements with opacity 0](https://github.com/michalsnik/aos/issues/397) all together when a user is navigating with their keyboard. A fix for this is to set the element with `opacity: 0.0001`. While this is not ideal, it guarantees the maximum number of screen readers will be able to read the link while not impacting design. In the future, when this ChromeVox bug is patched, opacity should be adjusted back to 0.

Unfocused link:
Screen Shot 2021-03-30 at 1 11 51 PM

Focused Link:
Screen Shot 2021-03-30 at 1 11 44 PM

Test with ChromeVox (audio) showing that selecting the Skip to Content skips the Nav and goes directly to focus on the "Welcome to Jam3."

https://user-images.githubusercontent.com/21203253/113030134-43c57e80-915b-11eb-8d6e-1af96990251a.mp4

Let me know if I should add the advisor warning to `adviserrc.json`.

**Additional comments:**

### DIFF
--- a/src/components/Nav/Nav.module.scss
+++ b/src/components/Nav/Nav.module.scss
@@ -16,6 +16,22 @@
     height: $navbar-height-desktop;
   }
 
+  .skipToContent {
+    position: absolute;
+    top: 0;
+    left: 0;
+    height: fit-content;
+    pointer-events: none;
+    opacity: 0.0001;
+  }
+
+  .skipToContent:focus,
+  .skipToContent:active {
+    color: $white;
+    background-color: $black;
+    opacity: 1;
+  }
+
   > .wrapper {
     display: flex;
     max-width: px(1440);
--- a/src/components/Nav/Nav.js
+++ b/src/components/Nav/Nav.js
@@ -25,6 +25,9 @@ function Nav() {
     <nav className={classnames(styles.Nav)}>
       <div className={styles.wrapper}>
         <ul className={styles.routes}>
+          <a tabIndex="0" aria-label="Skip to content" className={styles.skipToContent} href="#start-of-content">
+            Skip to content
+          </a>
           {Object.values(routes).map(({ path, title }) => (
             <li key={path}>
               <Link href={path}>
@@ -44,6 +47,7 @@ function Nav() {
           ))}
         </ul>
       </div>
+      <section aria-hidden="true" id="start-of-content"></section>
     </nav>
   );
 }


src/components/Nav/Nav.module.scss
```diff
@@ -16,6 +16,22 @@
     height: $navbar-height-desktop;
   }
 
+  .skipToContent {
+    position: absolute;
+    top: 0;
+    left: 0;
+    height: fit-content;
+    pointer-events: none;
+    opacity: 0.0001;
```
DonghyukJacobJang, line 25: can `opacity` just be 0?

DonghyukJacobJang (commented):
@KeshavChawla This is great to have! Thanks for your work!

src/components/Nav/Nav.module.scss
KeshavChawla: Hey @DonghyukJacobJang!

Normally I would say yes, but [ChromeVox has a bug where it skips reading out content with opacity 0 ](https://github.com/michalsnik/aos/issues/397) when a user is navigating with their keyboard. The hacky fix for this was to just set the element with `opacity: 0.0001;`. While not ideal, I think it guarantees the maximum number of screen readers will be able to read the link while not impacting design. That being said, if we don't need to support ChromeVox then I can set it to opacity 0.

jmckinnell: @DonghyukJacobJang Keshav was correct on this. For AAA sites any use of opacity animation has to use 0.0001 for it's starting opacity. So this is good to have. Let's approve and merge this if everything is groovy. ✌️

alemesa (approved):

alemesa:
Merging this @DonghyukJacobJang we tested this in Tonal and worked fine

PR #97, opened 2021-03-31
Feature: add proptype eslint

KeshavChawla:

- [ ] Code style update
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

**Adviser warns of an error with netmask (high).**

Advisory code is 1658: https://www.npmjs.com/advisories/1658**

## Solution Description

Replaces the Jam3 checkProps for React PropTypes. 

- [ ] N/A

**Additional comments:**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3488,12 +3488,6 @@
         "bowser": "2.1.0"
       }
     },
-    "@jam3/react-check-extra-props": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/@jam3/react-check-extra-props/-/react-check-extra-props-1.0.1.tgz",
-      "integrity": "sha512-oBnYpWE6vPDsH5HjQWNvEzG1wJmDCLTv/ue8prPk6S6e/JHEFF0nv0mDz6krVmxwtpcKh/O24xcg25HgtTRH/A==",
-      "dev": true
-    },
     "@jam3/react-hooks": {
       "version": "1.0.0",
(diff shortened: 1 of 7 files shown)


namniak:
proptype eslint is different from the `@jam3/react-check-extra-props` because eslint checks whether all props that are used inside the component code are specified in the props. while `react-check-extra-props` checks prop types + warns if some properties that are not specified in the props are passed down from outside. 
also `eslint` isn't smart enough on it's own as it works on a file-by-file basis, meaning if props are defind outside of the component e.g. id `data/prop-types` file eslist will not reference it correctly and won't warn about any mismatch in props defined vs props used as it would do if prop types were defined in the same file. But this is a very common use case where we define props and re-use them and therefore put in separate data files.

The bottom line is `@jam3/react-check-extra-props` can't be really replaced with this linter as they serve different purposes. Maybe in the future, a combination of it may work if we get both to work together nicely but so far I think `@jam3/react-check-extra-props` brings more value over eslint based on my experience in using both in different projects.

@DonghyukJacobJang I guess this PR can be closed ad there's no progress here except removing `@jam3/react-check-extra-props`? and if in future we combine the two that would be a new PR that has it's own challenges

edgardz:
@namniak @DonghyukJacobJang Related to https://github.com/Jam3/react-check-extra-props/pull/2

namniak:
@edgardz thanks! will check it out. Yes it'd be ideal to be able to use both together 🎉

neo:
I think once we have TypeScript, it would solve most of this problem for us without having to do any work :lolface:

PR #99, opened 2021-04-19
feature: integration with cicd, codeship

iranreyes:
### Features
- Add default CI/CD setup for Codeship Pro
- Reduce deployment times - with this setup 4min (but more improvements are coming => goal 1.5-2min)

### Changelog
- Add codeship pro files (env variables, .codehip folder, Dockerfile, codeship-services.yml, codeship-steps.yml)
- Add encrypt/decrypt scripts to help with the encrypted environment variables
- Add some basic documentation for setting up CI/CD
- Remove Lighthouse and WebHint from linters temporally, we are going to evaluate another way to do this without blocking deployments
- Move all the needed dependencies to run a build and linters to `dependencies` in the `package.json`, and keep the only development in `devDependencies`. (We started using `npm ci --only=production`)


namniak (approved):

neo:
Not sure if it's been done already, but if we want to improve the build time, we could consider caching the `.next/` folder across different builds.

PR #101, opened 2021-05-06
change: restructure Dockerfile and CICD pipeline

mdasilva:
Restructure Dockerfile to allow better use of image caching.
Restructure Codeship steps and yaml styling.
Add simple SAST scanning for develop. 

- [x] Code style update
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] N/A

**Aditional comments:**


Dockerfile
```diff
@@ -5,11 +5,13 @@ LABEL maintainer_architect="iran.reyes@jam3.com"
 # Set working directory
 WORKDIR /usr/src/app
 
+# Install dependencies
+COPY package-lock.json .
+COPY package.json .
+RUN npm ci
```
iranreyes, line 11: Noticed that you removed `--only=production`, any particular reason @mdasilva ?

codeship-services.yml
```diff
@@ -1,3 +1,9 @@
+njsscan-dev:
```
iranreyes, line 1: I like the example, however, it's better if we remove it from here because we are going to have a specific task for DevOpsSec Pipeline

Issue #103, opened 2021-05-20
[Feature] v10.2 upgrade (Webpack5)


namniak:
* adjust next config for Webpack5
* set up Storybook to work with Webpack5

references:
https://nextjs.org/blog/next-10-2
https://gist.github.com/shilman/8856ea1786dcd247139b47b270912324

PR #109, opened 2021-06-04
Upgrade Nextjs + storybook

namniak:
- React upgrade to v17
- NextJS upgrade to v10.2
- Storybook upgrade
- re-organized package json with a better division of deps vs dev deps

The main difference is that those upgrades require a major upgrade of Webpack from v4 to v5

Closes https://github.com/Jam3/nextjs-boilerplate/issues/103


DonghyukJacobJang (approved):

postcss.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  plugins: ['autoprefixer']
+};
```
neo, line 3: random question: why did we need to add a postcss config?

PR #113, opened 2021-06-22
[WIP] chore: test unsupported rewrite rule

alemesa:
Test out rewrite rule that came w/ 10.2 in Next.js this could replace the serverless approach

https://nextjs.org/blog/next-10-2#routing-based-on-headers-and-query-string-parameters


DonghyukJacobJang:
@alemesa closing this because this is for vercel 😢

PR #114, opened 2021-06-23
Next 11 Upgrade

alemesa:
Blog Release: https://nextjs.org/blog/next-11
Migration Guide: https://nextjs.org/docs/upgrading

Tested everything and seems ok, for once storybook didn't give me issues

Might be worth to add this too https://nextjs.org/docs/basic-features/eslint

Comes with Next 11 but throws some warnings that are about Vercel related stuff like the Image component


src/pages/_app.js
```diff
@@ -5,10 +5,10 @@ import '../styles/global.scss';
 
 import Layout from '../components/Layout/Layout';
 
-import detect, { isTouchDevice } from '../utils/detect';
+import detect, { isBrowser, isTouchDevice } from '../utils/detect';
 import { withRedux } from '../redux/with-redux';
 
-if (typeof window !== 'undefined') {
+if (isBrowser) {
```
namniak, line 11: `isBrowser` is evaluated in another file so this will become by default non-falsey and these chunks will make it into the server build.
the way Webpack works in the file by file. if the variable is set somewhere else it will know about it but it won't have access to its value therefore it is treated rather as `true` vs `false`

That's why I suggest people use these checks `typeof window !== 'undefined';` and `process.env.NODE_ENV === 'production'` directly setting them in detect or settings.

namniak: or `const isBrowser = typeof window !== 'undefined';` can b e put at the top of each file where it's used multiple times but not outside

namniak (changes requested):

DonghyukJacobJang (approved):

namniak (approved):

PR #121, opened 2021-06-25
Proposal: Update detect util

edgardz:

- [x] Code style update
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

## Problem Description

The current detect lib is not updated for 2 years and was lacking some new things we may need nowadays.

## Solution Description

This PR introduces a new and cleaner detect API with new features like:
- SSR fallback.
- Spoofed iPad detection.
- Improved in-app detection, including WeChat and TikTok.
- Low Power Mode detection on iOS.
- `useIsBrowser` hook to prevent hydration problems on mobile.

I changed all the references in the current code, but the things we bring from other projects, that rely on the old detect API, will have to be adjusted. It is fairly simple, though.

**Aditional comments:**


src/utils/detect.js
```diff
@@ -1,41 +1,186 @@
...
-  os: {},
-  bots: {}
-};
-
-if (isBrowser) {
-  detect.device = require('@jam3/detect').device;
-  detect.browser = require('@jam3/detect').browser;
-  detect.os = require('@jam3/detect').os;
-  detect.bots = require('@jam3/detect').bots;
+import bowser from 'bowser';
```
DonghyukJacobJang, line 1: The best place to improve detect util is [@jam3/detect](https://github.com/Jam3/detect) such as OS, Device and Browser. 
And also I was thinking to move the checking webp method as well.

package.json
```diff
@@ -51,11 +51,11 @@
     "extends": "react-app"
   },
   "dependencies": {
-    "@jam3/detect": "1.0.2",
```
DonghyukJacobJang, line 54: Instead of removing Jam3 detect library, we should improve and promote the util library.

examples/with-redux-ssr/_app.js
```diff
@@ -27,11 +28,8 @@ function App({ Component, pageProps }) {
         if (process.env.NODE_ENV !== 'production' && window.location.href.indexOf('?nostat') === -1) {
           require('@jam3/stats')();
         }
-        const { device, browser } = detect;
-        const classes = [device.isMobile ? 'mobile' : '', device.getType(), browser.getName()].filter((className) =>
-          Boolean(className)
-        );
-        document.body.className = [...document.body.className.split(' '), ...classes].filter(Boolean).join(' ');
+        const classes = [device.mobile ? 'mobile' : '', device.type, browser.name].filter(Boolean);
+        classes.forEach((c) => document.body.classList.add(kebabCase(c)));
```
DonghyukJacobJang, line 32: I am not 100% how this kebabCase is useful here. can we share some examples with me?

src/components/Image/Image.js
```diff
@@ -33,13 +35,17 @@ const Image = forwardRef(({ className, imageObj, loadingType, sizes }, ref) => {
...
     return { src, srcSet };
-  }, [isWebpSupported, file]);
+  }, [webpSupported, file]);
 
   const { desktopLg, desktopMd, desktopSm, tablet, mobile } = { ...DEFAULT_SIZES, ...sizes };
 
+  useEffect(() => {
+    check.webp('lossy', (supported) => setWebpSupported(supported));
+  }, []);
+
```
DonghyukJacobJang, line 48: In my PoV, the previous approach is better. The reason is that the hook will only run once when website is loaded (and no need to run and will not run). On the other hand, placing webp check hook here will run for every Image component at "componentDidMount" lifecycle which is not necessary.

src/pages/_app.js
```diff
@@ -26,11 +27,8 @@ function App({ Component, pageProps }) {
         require('@jam3/stats')();
       }
 
-      const { device, browser } = detect;
-      const classes = [isTouchDevice ? 'touch-device' : '', device.getType(), browser.getName()].filter((className) =>
-        Boolean(className)
-      );
-      document.body.className = [...document.body.className.split(' '), ...classes].filter(Boolean).join(' ');
+      const classes = [device.mobile ? 'mobile' : '', device.type, browser.name].filter(Boolean);
+      classes.forEach((c) => document.body.classList.add(kebabCase(c)));
```
DonghyukJacobJang, line 31: I am not 100% how this kebabCase is useful here. can we share some examples with me?

examples/with-redux-ssr/_app.js
edgardz: Before:
`Internet Explorer` => `internet explorer`
It ends up doing <body class="... internet explorer"> where "Internet" and "explorer" are added as 2 different classes.

With kebab:
`Internet Explorer` => `internet-explorer`
Each item in the classes array result in a single class added to the body, with a predictable structure: lower-case, dashes instead of spaces and so on.

This way we are ensuring any detect output will follow the same pattern when added as a body class.

src/components/Image/Image.js
edgardz: `check.webp()` caches the result and will only run the entire check once called for the first time. Subsequent calls will return the cached result straight away. No need to pollute the global state with something that will never change during the application execution.

src/pages/_app.js
edgardz: See https://github.com/Jam3/nextjs-boilerplate/pull/121#discussion_r659999842

src/utils/detect.js
edgardz: I'm not sure how other repositories depend on the detect util. And since it is a breaking change I thought it would be safer to introduce it here first, then improve it over a few projects, then promote it to the lib repo with a new major version.

package.json
edgardz: See: https://github.com/Jam3/nextjs-boilerplate/pull/121#discussion_r660004859

examples/with-redux-ssr/_app.js
DonghyukJacobJang: I see. Since Jam3 does not support IE starting from around 2020 January, we may not need this kebab case conversion if IE is the only case.

src/components/Image/Image.js
DonghyukJacobJang: I checked there is `_webp` in `Check` class. That is not what I am referencing. If Image component is instantiated from other components/pages, this hook will be called as many as Image component is instantiated in other places. For example, if Landing page has 9 Image components, this hook will run 9 times for the same result. 

That is one reason why it was saved in redux store instead of managing `webpSupported` locally. Even if this code is caching it to `_webp`, the callback will run it which can be avoided if we manage this globally.

edgardz: But... Wouldn't the same happen with the useSelector hook?
Also, another negative effect of the redux solution is the coupling it generates. Making it harder to test (on Jest, Storybook, etc) every single component that uses the Image component.

examples/with-redux-ssr/_app.js
edgardz: It will happen for every class now or in the future that might have a space in it. This change just make it even more future proof.

DonghyukJacobJang: If you think that, then I will have to suggest use existing API functions instead of `lodash.kebabCase`. 
```
classes.forEach((c) => document.body.classList.add(c.toLowerCase().replaceAll(' ', '-')));
```
No need to introduce 3rd party library for this.

edgardz: I thought it would be ok since we use external libs even for things like no-op...

src/components/Image/Image.js
DonghyukJacobJang: Of course, there are pros and cons for two different approaches.

**webpSupported in the redux store**

> Pros.
> - It will be only once setting `webpSupported` throughout the website lifespan. 

> Cons.
> - Image component will rely on `webpSupported` from the redux. (but there is a way around for this for Storybook thing)

**local `webpSupported`** 

> Pros.
> - It will be a standalone component.

> Cons. 
> - the hook will run and the callback method will run as many as Image component is being used.
> - if webp check returns `false` and calling callback `setWebpSupported` method, then all image components in the page will re-render. 

We can discuss further pros and cons to implement the best practice for this.

examples/with-redux-ssr/_app.js
DonghyukJacobJang: We are cool with 3rd party library if 
1) there is no alternative solution 
2) there is no time to find and code an alternative solution while development stage in projects
3) an alternative solution is more complex than the 3rd party library. 
4) vulnerability/security proof (some clients run an extensive security scan for the JS code. If 3rd party code has any level of a security flaw, then some client asks for a re-work)

that being said, it would not be always a good idea of adding 3rd party library, especially for the boilerplate project.

src/components/Image/Image.js
edgardz: This is how I see it:

**`webpSupported` in the redux store**

> Pros.
> - It will be only once setting `webpSupported` throughout the website lifespan. 

> Cons.
> - Image component will rely on `webpSupported` from the redux.
> - The `useSelector`, and all its internal logic, will run as many times as the Image component is being used.
> - Every test, on every project, for every component using the Image component will have to mock the redux store.
> - The execution cost of running a selector is higher than using simple React native hooks.

**local `webpSupported`** 

> Pros.
> - It will **also** be only once setting `webpSupported` throughout the website lifespan. 
> - It will be a standalone component.
> - Easier to test on every suite.
> - Easily understandable and portable.
> - Since it is simpler, the data comes almost directly from memory, not going through contexts, subscriptions, and reducers, it is less bug-prone.

> Cons. 
> - The native `useEfect` hook will run every time the component is instantiated, although the state won't change if the result is the same.

examples/with-redux-ssr/_app.js
edgardz: I've been using lodash on almost every project for years without any problem. It is almost an industry-standard. Also required by other packages that are already included in the boilerplate. Just look at the node_modules folder.
image

DonghyukJacobJang: I have to be against to the point that lodash is an industry-standard library. 

The lodash npm packages you shared above are not going to be a part of PROD JS bundles. Those are being used from dev dependencies. 

I agree lodash provides benefits, but there are more things to concern about several things when using it. I am very certain that you can find why I am saying this.

edgardz: I see. But at this point there is so many NOs to this PR I gave up and closed it.

src/components/Image/Image.js
DonghyukJacobJang: I like the point about being standalone. I think we can work on this further.

PR #124, opened 2021-07-12
fix: use NextJS default env local file name

DonghyukJacobJang:

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] Code style update
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## Problem Description

Using `BUILD_ENV` key and `.env.development` approach was introduced a long time ago before NextJS 9.4 release. 

## Solution Description

NextJS 9.4 + supports `.env.local` by default for the use of local environment. As a result, removing `BUILD_ENV` and renaming the file name.

reference: https://nextjs.org/docs/basic-features/environment-variables#loading-environment-variables

- [x] N/A

**Aditional comments:**


neo:
based on [this](https://nextjs.org/docs/basic-features/environment-variables#default-environment-variables), I think we might still want to use `.development` as the committed one? However, maybe we indeed don't need to specify `BUILD_ENV=development` in the dev npm script since it's the default

image

neo:
also on a kind of related note, I was thinking that maybe we want to bring back `npm start` rather than `npm run dev`? I personally still mistakenly run `npm start` from time to time

namniak:
> also on a kind of related note, I was thinking that maybe we want to bring back `npm start` rather than `npm run dev`? I personally still mistakenly run `npm start` from time to time

@neo I guess `npm start` would be valid rather for starting a prod node server (SSR) than local dev imho

neo:
then I'd just need to try harder getting used to it 😂  was just considering how infrequent we need to run that prod SSR server by hand...

edgardz:
I like the idea of having only these 2 envs:

.env.production (leveraging build time CI vars). Example:
```
NEXT_PUBLIC_ENVIRONMENT=$CI_ENV
```

.env.development (hard coding the CI vars we don't have while developing). Example:
```
NEXT_PUBLIC_ENVIRONMENT=local
```

This way we are able to do some cleanups:

- Remove this piece of logic from next.config.js:
```
require('dotenv').config({
  path: path.resolve(process.cwd(), `.env.${process.env.CI_ENV || process.env.NODE_ENV}`)
});
```

- Replacing the build scripts on package.json from
```
    "build:prod:static": "cross-env CI_ENV=prod next build && next export && cross-env CI_ENV=prod npm-run-all robots-txt-generator sitemap-generator",
    "build:staging:static": "cross-env CI_ENV=staging next build && next export && cross-env CI_ENV=prod npm-run-all sitemap-generator",
    "build:develop:static": "cross-env CI_ENV=develop next build && next export && cross-env CI_ENV=prod npm-run-all sitemap-generator",
```
to a single build script (bear in mind CI_ENV should be set on codeship-services.yml):
```
"build:static": "next build && next export && npm-run-all robots-txt-generator sitemap-generator",
```


It will also allow us to do conditional things in our code as well, like:
```
// hide GSAP warnings only in production, show on local, dev, staging...
gsap.config({ nullTargetWarn: process.env.NEXT_PUBLIC_ENVIRONMENT !== 'production' });
```

.codeship/artifact.sh
```diff
@@ -2,7 +2,9 @@
 set -e
 
 #### Running build ####
-npm run build:prod:static
+export COMMIT_ID=$(git rev-parse --short HEAD)
```
edgardz, line 5: We can also count the number of commits to have a numeric representation of the current version:
```
export COMMIT_NUMBER=$(git rev-list --no-merges --count HEAD)
```

.env.production
```diff
@@ -0,0 +1,10 @@
+# reserved env key
+BUNDLE_ANALYZE=false
+
+# custom env key
+NEXT_PUBLIC_WEBSITE_SITE_URL=$WEBSITE_SITE_URL
+NEXT_PUBLIC_ENVIRONMENT=$CI_ENV
+NEXT_PUBLIC_BUILD_ID=$COMMIT_ID
+NEXT_PUBLIC_GTM_ID=$GTM_ID
```
edgardz, line 8: Following my previous comment:
```
NEXT_PUBLIC_BUILD_NUMBER=$COMMIT_NUMBER
```

.codeship/artifact.sh
DonghyukJacobJang: I am looking for a better versioning way with major and minor numbers. In my opinion, knowing the exact commit id upfront would be better to keep track for now.

edgardz: I see. Considering the boilerplate itself, it makes sense to have proper versioning.
But for everyday projects, most of the time we just need to be able to compare if staging is up to date with development for QA purposes, so having a simple commit number in the console, for instance, would do the job.

edgardz (approved):

DonghyukJacobJang:
#126

PR #132, opened 2021-08-20
feature: implement TypeScript

DonghyukJacobJang:
CHANGELOG
===
1. Implement TypeScript 🚀 
2. Fix Storybook issue
3. Introduce Storybook doc (mdx)
4. Add useful Storybook addons


src/data/settings.ts
```diff
@@ -0,0 +1,6 @@
+// global
+export const resizeDebounceTime: number = 10; // in ms
```
neo, line 2: we don't have to specify the type for a variable (especially if it's a `const`) if it could be inferred by the assigned value

src/components/Layout/Layout.tsx
```diff
@@ -34,7 +32,7 @@ function Layout({ children }) {
   }, [router.events, handleRouteChange]);
 
   useEffect(() => {
-    checkWebpSupport('lossy', (isSupported) => dispatch(setIsWebpSupported(isSupported)));
+    checkWebpSupport('lossy', (isSupported: boolean) => dispatch(setIsWebpSupported(isSupported)));
```
neo, line 35: the boolean type might not be necessary if the `checkWebpSupport` is specific enough

src/pages/about/index.tsx
```diff
@@ -18,10 +20,6 @@ function About({ className }) {
   );
 }
 
-About.propTypes = checkProps({
-  className: PropTypes.string
-});
-
 About.defaultProps = {};
```
neo, line 23: I think we also wouldn't need `defaultProps` anymore as we could specify as default params in the function

neo (commented):
Very happy seeing this 😆  we are finally around the corner

.eslintrc.json
```diff
@@ -1,6 +1,7 @@
 {
-  "extends": ["react-app", "react-jam3", "plugin:@next/next/recommended"],
+  "extends": ["react-app", "react-jam3", "plugin:@next/next/recommended", "plugin:import/typescript"],
   "plugins": ["jam3"],
+  "parser": "@typescript-eslint/parser",
```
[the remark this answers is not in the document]
DonghyukJacobJang: I would keep it for now without this line, it throws errors.

tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "allowJs": true,
```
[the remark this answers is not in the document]
DonghyukJacobJang: definitely, it will happen. After providing TS 101 session, let's do it slowly. just trying to make sure everyone is feeling comfortable. We have already talked about this option in J3 open-source meeting, and they are aligned with it.

src/redux/index.ts
```diff
@@ -0,0 +1,40 @@
+import { TypedUseSelectorHook, useDispatch, useSelector } from 'react-redux';
+import { configureStore, createSlice, PayloadAction } from '@reduxjs/toolkit';
+
```
neo, line 3: My personal suggestion to follow a bit closer to this: https://github.com/Jam3/prj-2-089-xaxiscopilot/blob/develop/src/redux/index.ts, but they are more like code style and preference.

One thing I do want to bring up for discussion is that we might not need to export those types as they usually are not being used (or even worse there might be something not ideal/optimal when people are using those). Also would like to point out that `iAppState` and `AppState` would be essentially the same in our case?

```ts
type AppState = {
  prevRoute: string | null;
  isWebpSupported: boolean;
};

const initialState: AppState = {
  prevRoute: null,
  isWebpSupported: true
};

const { actions, reducer } = createSlice({
  name: 'app',
  initialState,
  reducers: {
    setPrevRoute(state, action: PayloadAction<string>) {
      return { ...state, prevRoute: action.payload };
    },
    setIsWebpSupported(state, action: PayloadAction<boolean>) {
      return { ...state, isWebpSupported: action.payload };
    }
  }
});

export const { setPrevRoute, setIsWebpSupported } = actions;

export const store = configureStore({ reducer });

// Use throughout your app instead of plain `useDispatch` and `useSelector`
export const useAppDispatch = () => useDispatch<typeof store.dispatch>();
export const useAppSelector: TypedUseSelectorHook<AppState> = useSelector;
```

neo (commented):
This should be my last batch of comments considering we would never be able to get it perfect in one go. So let's go let's go!!

(not saying it's not good enough 😆   on the contrast, it's looking very good to me!)

src/components/Image/Image.tsx
```diff
@@ -1,11 +1,19 @@
...
 import styles from './Image.module.scss';
 import sassVars from '../../styles/export-vars.module.scss';
 
+import { Breakpoints } from '../../utils/layout';
+import { useAppSelector } from '../../redux';
+
+type Props = {
+  className?: string;
+  imageObj: { file: string; alt: string };
+  loadingType?: 'lazy' | 'eager' | undefined;
```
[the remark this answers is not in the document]
DonghyukJacobJang: yes, this is a bit weird. you will be able to find why below.
```
Type '"lazy" | "eager" | "auto"' is not assignable to type '"lazy" | "eager" | undefined'.
  Type '"auto"' is not assignable to type '"lazy" | "eager" | undefined'.ts(2322)
```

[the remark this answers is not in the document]
neo: seems like it's right: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading
it doesn't take `auto` anymore, but I think we might still be able to remove `undefined` as it's specified with the question mark

package.json
```diff
@@ -24,14 +24,16 @@
     "component": "node scripts/component.js",
     "page": "node scripts/component.js --page",
     "\n ######################  Automation Tools  ######################": "",
+    "lint": "next lint",
     "lint-staged": "lint-staged",
-    "linters": "npm-run-all -p js-lint sass-lint adviser-dev ls-lint",
-    "linters-ci": "npm-run-all -p js-lint sass-lint adviser-ci ls-lint",
+    "linters": "npm-run-all -p lint js-lint sass-lint adviser-dev ls-lint tsc",
+    "linters-ci": "npm-run-all -p lint js-lint sass-lint adviser-ci ls-lint tsc",
```
neo, line 30: wondering do we need both lint and js-lint?

src/redux/index.ts
neo: I think defining the type of the state is still necessary because if one state is initially `null`, it wouldn't know what it normally could be.

neo (dismissed):
🚀 🚀 🚀

package.json
DonghyukJacobJang: It seems `next lint` a bit more strict. I am going to remove js-lint.

neo: Sounds great to me! :laughing:

PR #143, opened 2021-09-22
fix: remove postcss autoprefix

DonghyukJacobJang:
autoprefix works out of box, so removing the local postcss config


neo:
I think I know why it was there 🤦 
and it's not the first time we have similar problems...

🥁 

it's (trying to be) for storybook (although it wasn't working yet 😂 )

PR #152, opened 2021-09-29
Revert "fix: false positive for no-document-import-in-page"

DonghyukJacobJang:
Reverts Jam3/nextjs-boilerplate#151


DonghyukJacobJang:
if we have this eslint rule, we might lose track of the fix from NextJS. 
please find my additional comment from the link
https://github.com/Jam3/nextjs-boilerplate/pull/151#issuecomment-930532637

PR #161, opened 2021-10-27
[WIP] Next 12 upgrade

alemesa:
Blog Release: https://nextjs.org/blog/next-12
Migration Guide: https://nextjs.org/docs/upgrading

Next 12 swaps to a Rust/SWC compiler ditching Babel in the process. 
They will slowly support all babel config in the future but it will take time 
**Discussion in this thread** - https://github.com/vercel/next.js/discussions/30174

Builds would be faster and local development too 👇 

compare


.babelrc
```diff
@@ -1,8 +0,0 @@
-{
```
alemesa, line 1: @namniak since we're moving from Babel to Rust compiler, tagging you here just in case, cause I believe you add it this config file

namniak: @alemesa yeah this babel was brought in for v10.2 when they just introduced webpack 5 - kinda a patch. if the app can run without it now (and I presume it does) then we are good!

alemesa:
Let's hold a bit on this @DonghyukJacobJang I'm seen it a ton of issues being logged to the next repo related to v12, I think is better if we wait a bit while they iron out most issues

namniak:
@alemesa @DonghyukJacobJang we upgraded to v12 in one of the projects right after the release. so far so good (been about 2 weeks now but we are not at the completion yet to fully verify)

DonghyukJacobJang:
#193

PR #173, opened 2021-11-10
Update lock scroll

namniak:
* making `lock-scroll-body` as service as it works globally with the page


neo (dismissed):
"Previously tested in prod sites." sounds like it should be approved 😆

jmckinnell (dismissed):

DonghyukJacobJang (commented):
Thanks for arising the issue. One thing I would like to discuss is the code in `Layout.tsx`. 

The good thing about the code is that people does not need to worry about the bug related to that.
The downside is that it is an extra bundle if the project does not use `lock-body-scroll`, `get-scroll`  and `scroll-page`.

I am still thinking what will be the best implementation for us.

namniak:
> Thanks for arising the issue. One thing I would like to discuss is the code in `Layout.tsx`.
> 
> The good thing about the code is that people does not need to worry about the bug related to that. The downside is that it is an extra bundle if the project does not use `lock-body-scroll`, `get-scroll` and `scroll-page`.
> 
> I am still thinking what will be the best implementation for us.

fair enough - it's just something I like to add due to seeing this issue happening from time to time. def it's not necessary assuming people always clean up code on the components level.
I'm personally not too concerned about extra bundles in this case as this would be less than 0.5Kb in total but more about bloating code a bit.
I'll revert Layout

namniak:
@DonghyukJacobJang this can be turned in a `service` later on if we go with those

namniak:
@DonghyukJacobJang it's a service now

DonghyukJacobJang (approved):

PR #197, opened 2022-01-11
chore: upgrade Next + minor fixes

namniak:

- [x] Code style update
- [x] Build-related changes
- [x] Other, please describe: npm modules updates

- [ ] Yes
- [x] No

**Aditional comments:**
* upgrade Next to the latest stable version (fully back compatible) - fixes a few vulnerabilities
* update storybook
* adding `width: 100%` to body fixes content shift behind the overlays when lock body scroll is used
* removing `overflow-x: hidden;` from the body makes sure that `position: sticky` on the page elements will work - overflow set on any higher-level wrapper for any direction (x/y) will break it. Note that this is not a good practice to disable overflow on elements such as `html`, `body`, `main`, or other global app wrappers and issues with overflow should be fixed on the component level
* made sure react dev tools are disabled for production CI - no need to expose app logic


src/styles/global.scss
```diff
@@ -18,9 +18,9 @@ html {
 }
 
 body {
+  width: 100%;
```
alemesa, line 21: Singular Genomics 🤣  👍

alemesa (approved):

src/styles/global.scss
namniak: exactly!

PR #199, opened 2022-01-25
feature: storybook deployment [frontend]

alemesa:
**Frontend side PR**

This will deploy storybook on route `/storybook` for all our projects on all envs except `production`

---

To be merged in conjunction with **Backend side PR**: https://github.com/Jam3/frontend-devops-boilerplate/pull/78

We tested this in Noom Stories and it was working well, if you want to take a look:
- [Frontend Repo](https://github.com/Jam3/prj-2-227-noom-stories-devops)
- [Devops Repo](https://github.com/Jam3/prj-2-227-noom-stories/blob/develop/.codeship/artifact.sh#L16)

---

Lastly if we oficially migrate to CircleCI or Github Actions this is a pretty easy fix and will be added in the `build.sh` file for example: 

This is the students repo from this semester which is using GH actions instead of Codeship

https://github.com/Jam3/prj-3017-students-training-winter-2022/blob/develop/.github/scripts/build.sh


jmckinnell (commented):
I like it! Would love to hear @namniak or @iranreyes's thoughts from a BE perspective before we merge. But good from my end.

edgardz:
I like it. Just a few suggestions:

Rename the NPM script to `build:storybook` so we keep the naming pattern for building stuff:
```
    ...
    "build:ssr": "next build",
    "build:static": "next build && next export",
    "build:storybook": "build-storybook -c .storybook -s ./src -o ./out/storybook",
    ...
```
Also, note we can set the output to be inside the `./out/` folder by default since it is already ignored by git and will allow us to slightly simplify the artifact script:

```
...

npm run build:static

if [ "$CI_ENV" != "production" ]; then
  npm run build-storybook
fi

rm -rf /artifacts/build
mv ./out /artifacts/build

...
```

edgardz:
Also, I was not able to test it locally. What I did:

```
npm run build:static
npm run build:storybook
npx serve ./out
```

Acessing http://localhost:3000/storybook/ didn't work because Storybook was trying to load the scripts from `/`.
In order to solve this, I had to do a few things:

- Set the proper publicPath configuration on `.storybook/next-preset.js`:

```js
module.exports = {
  webpackFinal: async (baseConfig, options) => {
    ...
    if (process.env.NODE_ENV === 'production') {
      newConfig.output.publicPath = '/storybook/';
    }
    // Return the altered config
    return newConfig;
  },
  managerWebpack: async (config) => {
    if (process.env.NODE_ENV === 'production') {
      config.output.publicPath = '/storybook/';
    }
    return config;
  }
};

```

- Change the build script to:
```
    "build:storybook": "build-storybook -c .storybook -s ./src -o ./out/storybook -- --preview-url=/storybook/iframe.html",
```

- Disable `storybook-jira-addon` because for some reason we need to investigate, it breaks the build.

After doing this and rebuilding, I was able to access http://localhost:3000/storybook/

edgardz (approved):

PR #200, opened 2022-01-25
feature: improvements to typography

alemesa:
Probably a spicy PR but wanted to give it a shot

Ok couple of things I wanted to achieve here:

- organize everything typography related into 1 separate and single file only
- rename `font-slate-regular()` mixin to `font-body()` mixin to be more generic name, we always rename this to the project font usually in new projects anyways, usually a site is going to have a body font and a header font so I added a header font mixin too
- did a bit of a refactor on the font-size to allow for more values to be accepted under certain circumstances, is a bit hard to juggle for all possible outcomes but for example with these fixes all these options will work

```css
.div {
  @include font-size(16, 24, -0.1);
}

.div1 {
  @include font-size(20, 20, 3%, bold);
}

.div2 {
  @include font-size(20vw);
}

.div3 {
  @include font-size(20, 3%, 3%, bold);
}

.div4 {
  @include font-size(20, null, 1.5);
}

.div5 {
  @include font-size(16, normal, 3%, bold);
}

.div6 {
  @include font-size(20vw, normal, normal, bold);

  @include breakpoint(tablet) {
    @include font-size(80%, 1.5, 3%, 500);
  }

  @include breakpoint(desktopSm) {
    @include font-size(30, 60, 0.3, 500);
  }
}
```

Some other things:
- with these options you can also pass null as options, so in breakpoints you're not repeting yourself
- also I added font-weight as a 4th parameter


jmckinnell (commented):
I like the clean up of the typography system. Well done on that.

I don't like adding font weight to the font-size mixin. `font-weight` for me is something controlled by the typography mixin or set in the `@font-face` definition. Ideally once the typography system is setup a developer shouldn't have to worry about every setting `font-weight`, just using the correct typography mixin.

2. 
2.

src/styles/typography.scss
```diff
@@ -0,0 +1,99 @@
...
+
+// Define font-size, line-height and letter-spacing in function
+// @example scss Input
+//   p {
+//     // font-size: 16px;
+//     // letter-spacing: -0.1px;
+//     // line-height: 1.5; // This is line-height 24px and font-size: 16px
+//     @include font-size(16, 24, -0.1)
+//   }
+@mixin font-size($font-size, $line-height: null, $letter-spacing: null, $font-weight: null) {
```
edgardz, line 89: I like to have a central mixing for defining text properties.
But, since this defines other stuff besides the font-size I would rename it to `typography()` and add the font-family as a parameter as well:

```scss
// Define typography styles.
// We should avoid using this mixing directly in our class definitions.
// Instead, create repeatable typography mixins using this as the base.
// @example scss Input
//   @mixin typography-body {
//     @include typography('AdihausDIN', px(18), 1.2, -0.04em, 400);
//
//     @include media-desktop {
//        @include font-size(px(20));
//      }
//   }
//
//   @mixin typography-title {
//     @include typography('adineuePROCond', px(38), 0.894, null, 700);
//     text-transform: uppercase;
//
//     @include media-tablet {
//        @include font-size(px(108));
//      }
//   }
@mixin typography($font-family, $font-size, $line-height: null, $letter-spacing: null, $font-weight: null) {
  @include font-size($font-size);
  font-family: $font-family, #{$generic-fonts};
  line-height: line-height($font-size, $line-height);
  letter-spacing: letter-spacing($font-size, $letter-spacing);
  font-weight: $font-weight;
}
```

src/styles/typography.scss
```diff
@@ -0,0 +1,99 @@
...
+  @if math.is-unitless($font-size) {
+    font-size: px($font-size);
+  } @else {
+    font-size: $font-size;
+  }
```
edgardz, line 94: Note in the comment above I replaced this block with a font-size() mixin. This would allow us to use clamp() in order to prevent too small fonts on dynamic responsiveness (vw/rem based). Something like this:

```scss
// Sets font-size with a mininum and maximum size
// @example scss Input
//   p {
//     @include font-size(16vw, 12px);
//   }
@mixin font-size($font-size, $minimum-size: 14px, $maximum-size: $font-size) {
  @if math.is-unitless($font-size) {
    font-size: px($font-size);
    font-size: clamp(#{$minimum-size}, #{px($font-size)}, #{$maximum-size});
  } @else {
    font-size: $font-size;
    font-size: clamp(#{$minimum-size}, #{$font-size}, #{$maximum-size});
  }
}

```

src/styles/typography.scss
```diff
@@ -0,0 +1,99 @@
...
+
+/// Set Header Font as font family. This mixin's name is generic to abstract the font family name in case it changes
+/// @group Fonts
+/// @example scss Input
+///   .landing-section {
+///     @include font-header;
+///   }
+@mixin font-header() {
+  font-family: 'WorkSansSemiBold', #{$generic-fonts};
+}
```
edgardz, line 50: Based on the proposals above, these would have to be slightly refactored to something like:

```scss
@mixin typography-body {
   @include typography('WorkSansMedium', px(16), 1.2, -0.04em, 400);
}

@mixin typography-header {
  @include typography('WorkSansSemiBold', px(20), 0.8, null, 700);
}
```

src/styles/typography.scss
```diff
@@ -0,0 +1,99 @@
...
+
+/// Set Body Font as font family. This mixin's name is generic to abstract the font family name in case it changes
+/// @group Fonts
+/// @example scss Input
+///   .landing-section {
+///     @include font-body;
+///   }
+@mixin font-body() {
+  font-family: 'WorkSansMedium', #{$generic-fonts};
```
edgardz, line 39: This is a personal pet peeve. But I would like to move the $generic-fonts definition from the vars.scss to this file, cause it will never be used elsewhere.

alemesa: I agree, will move

src/styles/typography.scss
alemesa: I like it and will make these changes ⭐ we just to let everyone know about this change since everyone uses this mixin.

I also usually try to define mixins for all typography on the site in one place, but sometimes there are projects that have a ton of combinations and becomes really hard to unify it so I still have to use this mixin in components 😢

src/styles/typography.scss
alemesa: I actually like this a lot but got a question @edgardz didn't designers complain to you with this method, since font sizes will be slightly different in certain breakpoints in between (clamp)

src/styles/typography.scss
edgardz: Yeah. This is why it is important to stress out to the design team how important it is to componentize everything. Including typography. If we come to an agreement at the beginning of the project it will make everyone's lives way easier.

src/styles/typography.scss
edgardz: It really depends on the project. We can always set the minimum to 1px if it becomes an issue.

src/styles/typography.scss
jmckinnell: I think this is too much for 1 mixin. This means for each time I change the font size somewhere I have to specify the font family as well? I'd rather have 2 mixins. 1 for Family and weight and 1 for font size. If we want to create a mixin like this.

alemesa: I guess your could always pass null @jmckinnell 

```scss
@mixin typography($font-family, $font-size, $line-height: null, $letter-spacing: null, $font-weight: null) {
  @include font-size($font-size);
  font-family: $font-family, #{$generic-fonts};
  line-height: line-height($font-size, $line-height);
  letter-spacing: letter-spacing($font-size, $letter-spacing);
  font-weight: $font-weight;
}

@mixin h1() {
  @include typography(Arial, 10, 12, 1.2, -0.04em, 400);

  @include breakpoint(tablet) {
    @include typography(null, 12, 14);
  }

  @include breakpoint(desktop) {
    @include typography(null, 14, 16);
  }
}
```

but I guess what you're saying so my recommendation will be to have font-family as the last one, so we can skip the argument declaration for example like this:

```scss
@mixin typography($font-size, $line-height: null, $letter-spacing: null, $font-weight: null, $font-family: null) {
  @include font-size($font-size);

 // NOTE: I need to add a check here because $generic-fonts concatenation will force the rendering of this value otherwise
  @if($font-family) {
    font-family: $font-family, #{$generic-fonts};
  }
  line-height: line-height($font-size, $line-height);
  letter-spacing: letter-spacing($font-size, $letter-spacing);
  font-weight: $font-weight;
}

@mixin h1() {
  @include typography(10, 12, 1.2, -0.04em, 400, Arial);

  @include breakpoint(tablet) {
    @include typography(12, 14);
  }

  @include breakpoint(desktop) {
    @include typography(14, 16);
  }
}
```

At the end of the day I think the arguments should be in the order of what changes the most

cc: @edgardz

jmckinnell: @alemesa Sorry, I still disagree with combining font-size and font-family. If you want to create `typography` but simply have it make use of `font-size` that's fine. But I will continue to create specific mixins for each font family first as part of the project setup then create mixins for each typographic styles from the design bible. 

Large mixins/functions are bad design imo. 3 parameters max. It doesn't save you time, it just creates more things that you need to memorize.

src/styles/typography.scss
jmckinnell: @edgardz Not sure if this provides a benefit. Usually as part of the design review process we will point out when font sizes are too small the design team will give us updated sizes. In which case the size we enter into `font-size` will be the font we want.

src/styles/typography.scss
edgardz: @jmckinnell @alemesa The idea of the typography mixin is to be an umbrella mixin for defining all parameters of a specific font style. It is not supposed to be used outside typography.scss. This makes the fonts definitions easier to read and easier to port from Figma at the project setup phase, for instance:

```scss
@mixin typography-paragraph() {
  @include typography('Press_Start_2P', 16, 1.222, -0.04em, 400);

  @include media-tablet-portrait {
    @include font-size(px(20));
  }
}

@mixin typography-cta() {
  @include typography('Press_Start_2P', 20, 1.2, -0.02em, 500);
  text-transform: uppercase;
}

@mixin typography-h1() {
  @include typography('Press_Start_2P', 36, 1, -0.04em, 600);
  text-transform: uppercase;

  @include media-tablet-portrait {
    @include font-size(px(50));
  }
}

@mixin typography-h2() {
  @include typography('Press_Start_2P', 28, 1.136, -0.04em, 500);
  text-transform: uppercase;
}

```

Note it is only used at the base definition of the typography. Inside media queries, we just adjust what we need.

Now, if we need to change the font-size anywhere else we can simply use the font-size mixin or even define it directly like:

```scss
.someComponentClass {
  @include typography-cta;
  font-size: px(14);
}
```

This makes it clear we adapting some foundational typography. And if this starts to happen throughout the entire project it means we have a design problem and we have to chat with designers so they componentize the font styles.

src/styles/typography.scss
edgardz: @jmckinnell This is mostly a safeguard.

Let's consider a responsive approach where the font-size is based on the screen width.
Considering designers will only design for a specific mobile size (usually 375px wide), what happens when the website is opened on a 360px screen? All fonts originally designed to the minimum possible, let's say 10px, would be rendered even smaller. This usually happens with legal stuff, trademark copies, etc.

But yes, it is an edge case.
If we don't want to cover for it we can remove this logic and add it on a project basis depending on the need.

jmckinnell: @edgardz I see the benefit for sure, but I feel like this should be implemented on a case by case basis. Not something that we include in the boilerplate. Or possibly we can include it as an alternative mixin. i.e. `clamp-font-size` or something similar.

src/styles/typography.scss
jmckinnell: @edgardz I'm still not seeing the benefit of the `typography` mixin, especially if we're only using this at project setup. You don't save that much time and in my opinion makes the typography mixins harder to read. Really ask yourself which is easier to understand at a quick glance. Typing out `font-family` or `font-weight` takes no time at all with auto complete in Code so that argument for easier to type isn't there.

```
@mixin typography-paragraph() {
  @include typography('Press_Start_2P', 16, 1.222, -0.04em, 400);

  @include media-tablet-portrait {
    @include font-size(px(20));
  }
}
```
or
```
@mixin typography-paragraph() {
  font-family: 'Press_Start_2P';
  font-weight: 400;
  @include font-size(16, 18, -0.04);

  @include media-tablet-portrait {
    @include font-size(20);
  }
}
```

alemesa:
@edgardz @jmckinnell trying to find a middle ground here:
- we can keep the font-size mixin as it is w/ the improvements from this PR
- then I'll add the typography mixin as Edgard says

the team or lead dev can choose what to do which approach to use

either this:

```scss
@mixin typography-body {
   @include typography('WorkSansMedium', px(16), 1.2, -0.04em, 400);
}
```

or this:

```scss
@mixin typography-body {
  @include font-size(16, 1.2, -0.04em);
  font-weight: 400;
  font-family: 'WorkSansMedium'
}
```

or go wild

---

I would also flip the order of the typography if is cool too make it cleaner and avoid passing null to arguments, since font family usually doesn't change between breakpoints

so:

```scss
@mixin typography-body {
   @include typography(px(16), 1.2, -0.04em, 400, 'WorkSansMedium');
}
```

---

Last notes:
- personally I would try on my projects to push the designers to give me everything before hand and respect a typography system and I will ignore when they break their system, because I believe it will make our life easiers in the long run
- I don't mind big mixins, I actually think they add structure and force people to be declarative, because in theory with autocomplete, etc I could say the same about every mixin, but I know some devs hate them, so I'm probably not in the majority
- sorry, I knew this PR was gonna be spicy, is your fault @neo

edgardz:
@alemesa A quick note about mixins parameters is we don't have to write down all parameters if we don't want to. Sass allows us to set only the ones we want, like:

````scss
@include typography($font-size: px(20), $font-family: 'WorkSansMedium');
```

Note we don't have to follow a specific order nor add null for the missing parameters.

src/styles/typography.scss
edgardz: @alemesa @jmckinnell
Typography mixin is a personal preference actually. So no real hard reason to keep it for everyone.

We can get rid of it and only keep/create mixins for related parameters.

Now, since we have one mixin related to the size and spacings of the text, we could have another one related to the font selection (family, weight, style). What do you think?

```scss
@mixin typography-paragraph() {
  @include font-family('Press_Start_2P', 400, italic);
  @include font-size(16, 18, -0.04);

  @include media-tablet-portrait {
    @include font-size(20);
  }
}
```

edgardz (approved):

jmckinnell (approved):

neo:
(a year late to the game 🤣)

I agree with Jon that adding a mixin trying to cover _everything_ seems a bit unnecessary, and I'm glad that we came to agreement that instead of replacing the old one, we add this new one to be able to do composition. I like even more Edgard's other suggestion of having another mixin for "font family" or "font variant", and then we can compose with the two.

Additionally, in the spirit of having the new mixin only available in the file, I think we can utilize the (relatively) new sass module system (in this case and more generally); specifically, a [private](https://sass-lang.com/documentation/at-rules/use#private-members) mixin will help us in our case.

PR #201, opened 2022-01-25
feature: gsap init + improvements [proposal]

alemesa:
This PR main goal is too add `gsap-init` which will fix a couple important things:
- storybook where animations won't work even if we import gsap in our components
- `scroll-page` util; because this util which is used by the lock service is using `ScrollToPlugin` from gsap right now by default doesn't work since we're not importing this anywhere, therefore vertical positioning is not getting restored.

Then some more proposals:
- Add a bit of motion as default on our pages, so for example coops have an idea of how we do motion from the beginning (but I could see how this might be too much and is better to keep the generator clean)
- Then probably the most spicy `registerEffect` I think we as a company we should start to use this https://greensock.com/docs/v3/GSAP/gsap.registerEffect() since we repeat ourselves sometimes a lot in projects when it comes to animation, things like fade from the bottom animations usually in motion bibles are all the same, and registerEffect allows for variations anyways

Personally I have only used registerEffect in personal projects and coops projects so I don't have a production example that it works perfect but it should and also the greensock overlords talked about this in the presentation that they did here.

### DIFF
--- a/src/utils/gsap-init.ts
+++ b/src/utils/gsap-init.ts
@@ -1,0 +1,27 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import gsap from 'gsap';
+import ScrollToPlugin from 'gsap/dist/ScrollToPlugin';
+import ScrollTrigger from 'gsap/dist/ScrollTrigger';
+
+function gsapInit() {
+  gsap.registerPlugin(ScrollToPlugin, ScrollTrigger);
+  gsap.defaults({ ease: 'power2.out', duration: 0.333 });
+  gsap.config({ nullTargetWarn: false });
+
+  gsap.registerEffect({
+    name: 'fadeIn',
+    extendTimeline: true,
+    effect: (targets: any, config: any) => {
+      return gsap.from(targets, {
+        duration: config.duration,
+        autoAlpha: 0,
+        y: config.y,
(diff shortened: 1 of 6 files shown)


src/utils/gsap-init.ts
```diff
@@ -0,0 +1,11 @@
+import gsap from 'gsap';
+import ScrollToPlugin from 'gsap/dist/ScrollToPlugin';
+import ScrollTrigger from 'gsap/dist/ScrollTrigger';
+
+function gsapInit() {
+  gsap.registerPlugin(ScrollToPlugin, ScrollTrigger);
+  gsap.defaults({ ease: 'power2.out', duration: 0.333 });
```
jmckinnell, line 7: What no expo!? 😛 jk

jmckinnell (dismissed):
This looks good. nice and simple.

src/utils/gsap-init.ts
alemesa: haha I like `expo` way more too 😄 but I tried to play it safe with `power2`

jmckinnell (approved):

src/utils/gsap-init.ts
neo: I feel like most of them are actually expo lol

PR #202, opened 2022-01-28
feature: automatic eslint import sort

alemesa:
Addding automatic import sort so we don't need to do this manually, if you got eslint and prettier install will sort everything in the order defined in the [eslint config file](https://github.com/Jam3/nextjs-boilerplate/compare/main...feature/automatic-import-sort#diff-6884918dc8291219be508e05e28965b958c734def85324f3b53858ea4702090f), I try to follow the order that we have in our generator right now.

Plugin will also order alphabetically between groups

Open to hear suggestions about proper order, etc

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,7 +7,7 @@
     "plugin:import/typescript",
     "prettier"
   ],
-  "plugins": ["jam3"],
+  "plugins": ["jam3", "simple-import-sort"],
   "parser": "@typescript-eslint/parser",
   "settings": {
     "react": {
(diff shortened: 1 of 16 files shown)


namniak:
@alemesa I love it! 
Anything that is pre-defined and does things automatically to make code look uniform is great!

namniak (approved):

jmckinnell (approved):
I like this as well. But it will be doubly important for us to make sure FL has Eslinit inistalled in their VSCodee. Otherwise it's going to be a bad time for the rest of us.

neo (approved):
If both Vadim and Jon like something, it's gotta be really really good 😆 so I'll just follow and approve as well!

(I'm guessing it will also solve my problem that vscode auto import always goes to the bottom)

Good one! Thank you!! ❤️

PR #204, opened 2022-01-28
feature: register effect gsap

alemesa:
Proposal:
- Add a bit of motion as default on our pages, so for example coops have an idea of how we do motion from the beginning (but I could see how this might be too much and is better to keep the generator clean)
- Then probably the most spicy `registerEffect` I think we as a company we should start to use this https://greensock.com/docs/v3/GSAP/gsap.registerEffect() since we repeat ourselves  sometimes a lot in projects when it comes to animation (including me), things like fade from the bottom animations usually in motion bibles are all the same, and registerEffect allows for variations anyway

The not repeating ourselves, is inspired by the Singular Genomics motion bible which had like only 2-3 types of animation repeated everywhere 😆 but I still added `.from autoalpha: 0, y: 50` 100 times

Personally I have only used registerEffect in personal projects and coops projects so I don't have a production example that it works perfect but it should and also the greensock creators talked about this in the presentation that they did here.

### DIFF
--- a/src/pages/about/index.tsx
+++ b/src/pages/about/index.tsx
@@ -1,5 +1,6 @@
-import { memo, useRef } from 'react';
+import { memo, useRef, useEffect } from 'react';
 import classnames from 'classnames';
+import gsap from 'gsap';
 
 import styles from './index.module.scss';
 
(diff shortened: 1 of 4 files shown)


src/pages/about/index.tsx
```diff
@@ -11,11 +12,23 @@ type Props = {
 
 function About({ className }: Props) {
   const containerRef = useRef<HTMLElement>(null);
+  const headerRef = useRef<HTMLHeadingElement>(null);
+
+  useEffect(() => {
+    const headerElRef = headerRef.current;
+    gsap.effects.fadeIn(headerElRef, { delay: 0.1 });
```
jmckinnell, line 19: If we don't use a timeline here theen there will be a blip of the header showing then animating in correct?

jmckinnell (commented):
I like it! One comment on the Nav animation.

jmckinnell (approved):

edgardz (approved):

PR #207, opened 2022-02-01
feature: improve gsap.registerEffect to be more TypeScript friendly

neo:
This will help most significantly in applying effects to provide type checking as well as auto-complete. It will also help with registering the effects on defining required/optional config. The con being ppl adding an effect would need to go to two different files to achieve that...


alemesa (approved):
looks good to me, would love if @edgardz takes a quick look or anyone else with more TS experience than me

edgardz (approved):

PR #208, opened 2022-02-08
Circle CI Migration

alemesa:
Waiting on Iran's blessing


MLCochrane:
Two quick questions, but mostly because I don't know the answers:

1. If we're attaching a previous job's workspace and then checking out right after, does that potentially overwrite changes we're carrying over? I can't remember.
2. Should the build job depend on the `linters` job as well or does it matter? I think they'd run in parallel so the linters job could potentially fail but the build is still deployed.

alemesa:
> Two quick questions, but mostly because I don't know the answers:
> 
> 1. If we're attaching a previous job's workspace and then checking out right after, does that potentially overwrite changes we're carrying over? I can't remember.
> 2. Should the build job depend on the `linters` job as well or does it matter? I think they'd run in parallel so the linters job could potentially fail but the build is still deployed.

hey @MLCochrane 
1. I think checking out is like github actions, it will clone the repo or if there is a repo already will do nothing but I could be wrong - I think I tried without the checkout and didn't work, told me that files didn't exist - I can give you access to my personal repo and circle ci account if you want to give it a look since we still don't have an organization account here

2. linters don't need a build process since is just code review linter, but actually you're right about one failing and the other one succeeding, I think the linters don't need to actually run in deployment branches and it will avoid this conflict.

something like this, what do you think?

```
      - linters:
            branches:
                ignore:
                    - main       
            requires:
                - prepare
          
```

MLCochrane:
Hey @alemesa 
> 1. I think checking out is like github actions, it will clone the repo or if there is a repo already will do nothing

So I did a quick test, and looks like it will overwrite changes we made to existing files but won't affect any new ones (if that makes sense). I have a stupid example [here](https://app.circleci.com/pipelines/github/MLCochrane/circ/5/workflows/fec43d8e-2165-40ca-a225-61a1d727caa0) where we update a text file, attach the workspace to the next job, then checkout. Looks like in the `deploy` job the first "get text content" step shows our changes (second "New content" line) but after we checkout it shows what we had originally in the repo. Listing out the files before and after shows the same aka. new file wasn't affected.

Don't think that'd actually affect anything here because node_modules and final build files wouldn't be in the repo, but maybe something we'd run into if we needed to modify a config file in the repo before deploying or something? I'm not creative enough to think of where it'd actually be an issue, but something to maybe be mindful of if there are weird issues down the line.

> 2. linters don't need a build process since is just code review linter

Makes sense! Having that job ignore the main branch seems like a reasonable solution. Once again probably wouldn't end up being an issue since it's a) just for code standards and b) would have already been run on the changes before it made it to that branch anyways. Could always split up the deploy and build into separate jobs if we had other tests that should block deployment (testing and building would happen in parallel) but no need here.

alemesa:
Amazing @MLCochrane ⭐ 
1. gonna test in my personal repo one more time without the checkout but I think it shouldn't affect like you said anwyays (I gave you access to my repo https://github.com/alemesa/aws-testing in case you want to take a look)
2. will update with this change 👍 

also got a question how would you handle multiple envs? best practices? I could just repeat it in the same file tbh but maybe there is a smarter way

I need to add develop and staging workflow probably, they will also have different env variables

MLCochrane:
> also got a question how would you handle multiple envs? best practices? I could just repeat it in the same file tbh but maybe there is a smarter way
> 
> I need to add develop and staging workflow probably, they will also have different env variables

@alemesa This is definitely something that bugged at first as I felt I was repeating myself too much. I used to have almost identical jobs (deploy-staging, deploy-prod, etc.) with just different ENV variables and then tried moving as much duplicate yaml into "groups" or whatever they're called like you've done with the "defaults." I was just looking to [Contexts](https://circleci.com/docs/2.0/contexts/) which looks more promising.

Basically organization-wide groups of variables, but you can set different contexts for each job in the workflows. So, if for example the only thing we want to change is ENV variables for deploying to separate buckets, servers, etc. we could use _one_ job and then just add a different context and filter on the branch.

Modified my silly repo from above with that: https://github.com/MLCochrane/circ/blob/main/.circleci/config.yml

So `deploy` is added twice in the workflows with just a different `name`, `context` and branch filter. There are two contexts within CircleCI that both have the same variable `$SECRET_STRING` but with different values. Don't know if you can see artifacts or not, but the new-file.txt shows the two different variable values: [production](https://app.circleci.com/pipelines/github/MLCochrane/circ/9/workflows/53cca5be-2cd9-4064-845b-d2e2d0c66d59/jobs/19/artifacts) and [staging](https://app.circleci.com/pipelines/github/MLCochrane/circ/8/workflows/700f25dc-c1b4-45c8-89ec-5a8396d2aae9/jobs/17/artifacts).

This _feels_ better than what I've done in the past, but I've only just looked into it so there might be some "gotcha" we don't know about.

alemesa:
really good info @MLCochrane will look into this 👍

iranreyes:
Will take a look to this today

alemesa:
@MLCochrane took a look at contexts and is actually really nice, but I believe is more for shareables envs across organizations but not really for the same project

In my ideal world this works like Gh actions where you declare the stuff that changes between envs for example
```
context: 

 develop: 
  AWS_KEY: AWS_KEY_DEV
  CLOUDFRONT: CLOUDFRONT_DEV

 staging:
  AWS_KEY: AWS_KEY_STAGING
  CLOUDFRONT: CLOUDFRONT_STAGING
```

then I can use that below depending on the branch for example

alemesa:
I kinda figured it out @MLCochrane I think the consensus is a script that declares different variables per branch:

Added it [here](https://github.com/Jam3/nextjs-boilerplate/pull/208/files#diff-4c2d441a2952186fa4962027a53901ad8adc024f7f1d37892344a762e1d815ad)

Did some extra cleanup and splitted everything into external scripts.

I got all this running with the same setup in my local repo, you both got access:
- [circle ci workflow](https://app.circleci.com/pipelines/github/alemesa/aws-testing/32/workflows/39e3c68c-c7bb-4bfd-b17c-1de461d6efa9)
- [repo](https://github.com/alemesa/aws-testing/tree/main/.circleci)

I still prefer Github actions 😄 

cc: @iranreyes

MLCochrane:
@alemesa Cool - moving to a script is probably cleaner than any other options I can think of right now.

I think technically setting ENV variables in the job means CircleCI won't hide them if they get echoed out for example, _BUT_ if a malicious 3rd party is able to run arbitrary jobs on a repo then we've probably got bigger problems to worry about than just one project's secrets. Plus we already saw you could just save them to an artifact so hiding in a job doesn't do much anyways?

> I still prefer Github actions 😄

Haha still good to test what's out there! Maybe GH Actions is better suited for most of the projects we have anyways.

iranreyes:
Blessed

MLCochrane:
@alemesa @iranreyes  I have one final thought regarding the scripts which I think would affect both CircleCi or Github Actions. I realize some of the scripts were updated for Codeship first so it's not totally relevant, but I there might be some time savings down the line by ensuring we split up certain commands into separate jobs rather than keeping in a script.

For example, in the build file we have two npm scripts that run one after another just based on an ENV variable. 
```bash
# /scripts/build.sh
npm run build:static

if [ "$CI_ENV" != "production" ]; then
  npm run build:storybook
fi
```
This would mean that when both need to run the 2nd has to wait for the 1st to complete rather than running in parallel. It's been so long since I've used to storybook to remember if it needs a build, but something to keep in mind if we add more scripts as part of the "build."

Might not be a ton of time saved per workflow, but if something like this runs on every push outside of production then that will add up. The alternative would be splitting any scripts that don't depend on one another into separate jobs so they can be run at the same time in different containers (ex. separate "build" and "storybook" jobs). Pretty sure CircleCI and GH Actions handle this pretty similarly.

PR #218, opened 2022-03-23
Organize and cleanup Storybook implementation

edgardz:
Jira Ticket ID: <JIRA Ticket ID> 

- [ ] Code style update
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

## Problem Description

Storybook implementation was not optimal.

## Solution Description

- Removed unused JIRA plugin (it was breaking the build when enabled).
- Cleaned up storybook bootstrap implementation.
- Added a checkerboard background to easily identify transparencies.
- Moved documentation stories out of components folder.
- Added some preset colored backgrounds so we can test our components in different scenarios.
- Added some preset screen sizes so we can test responsivity.
- Normalized both Next and Storybook body classes in order to avoid different results when these classes are used.
- Added a new sass util for better integration with sass variables (See [2.Colors.stories.mdx](https://github.com/Jam3/nextjs-boilerplate/compare/feature/storybook-cleanup?expand=1#diff-906ed2ef6b533b6a9ae72eceb8a03ed469d2f93f054af5e2f24d143555f2b166) for an example)

- [ ] N/A

**Aditional comments:**


src/utils/sass.ts
```diff
@@ -0,0 +1,17 @@
...
+    .reduce((obj, key) => {
+      obj[key.split('-')[1]] = sassVars[key];
+      return obj;
+    }, {} as { [key: string]: string });
+}
+
+export default {
+  layout: getVariables('layout'),
+  color: getVariables('color')
+};
```
alemesa, line 17: I think it would be nice if we have an `all` properties too in case we have isolated cases

```suggestion
export default {
  layout: getVariables('layout'),
  color: getVariables('color'),
  all: sassVars
};
```

alemesa (dismissed):
Looks amazing @edgardz !! 🌟 Thank you

I think the only thing that I would recommend is moving the design system stories (intro, colors, typography, svg) inside the `.storybook` folder itself, imo feels weird that they are inside `docs` for me that is more for .md onboarding files, legal, etc

also left a small comment about the sassVars

alemesa (approved):
🔥

PR #224, opened 2022-04-22
Improve services intellisense

edgardz:

- [ ] Code style update
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## Problem Description

Services autocompletion/autoimport were not working well.

## Solution Description

Creating a variable pointing to the service instance solved the intellisense issues.

- [x] N/A

**Aditional comments:**
- Fixed the resize service with the latest code used in a real project.
- Also added 2 new services for handling cookies and local-storage. By using them instead of directly interacting with the browser we are able to listen for changes and react to them.

### DIFF
--- a/src/services/cookie.ts
+++ b/src/services/cookie.ts
@@ -1,0 +1,28 @@
+import Cookies from 'js-cookie';
+
+type CookieListener = (name: string, value: string | undefined) => void;
+
+class Service {
+  listeners: CookieListener[] = [];
+
+  set = (name: string, value: string, options: Cookies.CookieAttributes = { expires: 30 }) => {
+    Cookies.set(name, value, options);
+    this.listeners.forEach((listener) => listener(name, value));
+  };
+
+  get = (name: string) => {
+    return Cookies.get(name);
+  };
+
+  listen = (listener: CookieListener) => {
+    if (!this.listeners.includes(listener)) this.listeners.push(listener);
+  };
+
+  dismiss = (listener: CookieListener) => {
+    this.listeners = this.listeners.filter((l) => l !== listener);
+  };
+}
+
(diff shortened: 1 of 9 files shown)


src/services/resize.ts
```diff
@@ -2,31 +2,41 @@ import { device } from '@jam3/detect';
...
+    clearTimeout(timeout);
+
+    timeout = setTimeout(() => {
+      this.listeners.forEach((listener) => listener(e));
+      if (device.mobile) {
+        timeout = setTimeout(() => {
+          this.listeners.forEach((listener) => listener(e));
+        }, 500); // some mobile browsers only update window dimensions when the rotate animation finishes
+      }
+    }, this.debounceTime);
```
namniak, line 20: use `resizeDebounceTime` from `settings` instead as that value is also used in other places like `src/hooks/use-layout.ts` and we should keep it centralized

src/services/cookie.ts
```diff
@@ -0,0 +1,28 @@
...
+  };
+
+  dismiss = (listener: CookieListener) => {
+    this.listeners = this.listeners.filter((l) => l !== listener);
+  };
+}
+
+const CookieService = new Service();
+
+export default CookieService;
```
namniak, line 28: this is great.
with this added, we might want to make use of it right away within https://github.com/Jam3/nextjs-boilerplate/blob/main/src/hooks/use-cookie-banner.ts

edgardz: I have an upcoming PR targeting just the cookie banner once this one is merged.

namniak (approved):

jmckinnell (approved):

PR #225, opened 2022-04-27
Simplify cookie banner and Analytics implementation

edgardz:

- [ ] Code style update
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## Problem Description

Cookie banner and analytics logic were spread through several files and folders, responsibilities were fragmented, and the Layout component was used to couple both functionalities using an unnecessary hook, causing unnecessary layout-level re-renders and making it difficult to maintain.

## Solution Description

Centralize cookie banner logic on its own component. Centralize analytics logic on its own service. Use the generic cookie service to bridge the communication between both.

- [x] N/A

**Aditional comments:**


DonghyukJacobJang (commented):
Thanks for improving the cookie banner and analytics. I love how you deal with the events between them. 

One thing I wanted to suggest is to remove `log`;
- If you already confirmed (I also tested) that GTM kicks in whenever consent cookie, then it is just an extra code for nothing.
- there are alternatives to `log` in the trackEvent function by adding chrome extensions.

PR #226, opened 2022-04-27
A bunch of hooks

edgardz:

- [ ] Code style update
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

## Problem Description

We often need a set of hooks on every project and end up copy/pasting them from previous projects.

## Solution Description

I'm adding the most common ones I've been copying/pasting over the last year. I like to have these included on the boilerplate itself, instead of in "@jam3/react-hooks" because it allows us to make modifications based on the project requirements, which often becomes very complicated if the hooks are defined in a separate package.

- [x] N/A

**Aditional comments:**

### DIFF
--- a/src/hooks/use-combined-refs.ts
+++ b/src/hooks/use-combined-refs.ts
@@ -1,0 +1,24 @@
+import { ForwardedRef, MutableRefObject, useEffect, useRef } from 'react';
+
+// https://itnext.io/reusing-the-ref-from-forwardref-with-react-hooks-4ce9df693dd
+
+function useCombinedRefs<T>(
+  ...refs: (MutableRefObject<T> | ForwardedRef<T> | ((r: T) => void))[]
+): MutableRefObject<T> {
+  const targetRef = useRef<T>();
+
+  useEffect(() => {
+    refs.forEach((ref) => {
+      if (!ref || !targetRef.current) return;
+      if (typeof ref === 'function') {
+        ref(targetRef.current);
+      } else {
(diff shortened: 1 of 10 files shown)


namniak (commented):
@edgardz instead of adding these here we should add them instead in the separate repo here https://github.com/Jam3/react-hooks
it's already used with generator

jmckinnell:
> @edgardz instead of adding these here we should add them instead in the separate repo here https://github.com/Jam3/react-hooks it's already used with generator

Actually looking at these. I'd say that most of them should sit within the boilerplate because they use the new services that have been added. If they sit in the react-hooks repo then they can't make use of the services.

namniak:
we also have some of them already in the hooks repo such as intersection observer, orientation etc
I think we still need to find a way to keep this modular in the hooks repo rather than bloat generator as these are not all necessarily needed on every project

alemesa:
@jmckinnell @namniak I think we can talk more in the open source meeting tomorrow

I have been working on updating the hooks repo to a **2.0** with **Typescript**, way **more hooks** and multiple **improvements** (I'm also using the **services** heavily)

I already talked with @edgardz about it and we agree to keep it in the repo so that the hooks are exposed, since sometimes we have to change them slightly but at the same time is true that is a bit of bloat and we don't need all of them usually.

Regardless I would like to update the hooks repo since is the perfect case of a modular library and can be used outside the generator.

I'm a bit busy now with Complex but I'll try my best to have it next week

PR #227, opened 2022-05-12
[Feature] Responsive Image Component (Serverless Image Handler)

grez96:
This PR introduces a front-end component that can dynamically process images by changing their size, fitment, crop, and color.

**NOTE**: `ResponsiveImageComponent` relies on infrastructure that is deployed as part of [Jam3/frontend-devops-boilerplate](https://github.com/Jam3/frontend-devops-boilerplate). Without supporting infrastructure created by the AWS Serverless Image Handler solution, this component will fallback to using local image without any dynamic processing.

Since this PR relies on infrastructure to test functionality properly, I have specifically created a temporary deployment that mimics a typical Jam3 production static site just for testing this PR.
Front-End Repository: [Jam3/glebs-frontend-sandbox](https://github.com/Jam3/glebs-frontend-sandbox)
Dev-Ops Repository: [Jam3/glebs-frontend-sandbox](https://github.com/Jam3/glebs-frontend-sandbox-devops)
URL: https://prod.glebs-frontend-sandbox.jam3.cloud.jam3.net/

1. Begin by removing the image hash that is generated by Next.js image optimization. This is a requirement because the hashed names generated by Next.js causes a conflict in the retrieval of images from S3. (Potentially we can entirely remove compile time image optimization, but keeping it for safety and backwards compatibility).
https://github.com/Jam3/nextjs-boilerplate/blob/236136a2f59932c4ac04b910f6b768af25297f35/next.config.js#L10

2. There is a new environment variables that must be specified `NEXT_PUBLIC_SERVERLESS_IMAGE_HANDLER_CDN` (the URL of deployment environment with `/images/` suffix). The value for this variable will be provided by the dev-ops or TA assigned on project. For an example reference [Jam3/glebs-frontend-sandbox](https://github.com/Jam3/glebs-frontend-sandbox).
https://github.com/Jam3/nextjs-boilerplate/blob/0615ea974f5a2298a9582dcd715fb028c40f4ca5/.env.development#L18-L19

3. Whitelist the image CDN URL for the `img-src` CSP. This is only required for local development, once deployed, the image CDN should live on the same origin as the application.
https://github.com/Jam3/nextjs-boilerplate/blob/01ebcaa0765bf49d0b95a8ea09ffde4f419d19fc/src/components/Head/ContentSecurityPolicy.tsx#L25-L26

4. The remaining files define the actual `ResponsiveImage` component. The main functionality of `ResponsiveImage.tsx` is to process a local image src into an encoded URL string expected by the CDN. Also there is direct support for `srcset` and `sizes` attributes.

**NOTE** The AWS services actually runs processing on images **already** uploaded to an S3 bucket. So this component will produce a 404 error if the image that you are trying to optimize has not been deployed yet as part of a previous build.

**NOTE** Unfortunately this will not work correctly with Storybook. It seems like Storybook runs some custom WebPack processing changing the name of the image and no longer matches what is on S3

Additional Readings: https://aws.amazon.com/solutions/implementations/serverless-image-handler/


src/components/ResponsiveImage/ResponsiveImage.tsx
```diff
@@ -0,0 +1,104 @@
...
+  ) {
+    return localSrc;
+  }
+
+  // API expects no leading slash
+  const localSrcNoLeadingSlash = localSrc.replace(/\//, '');
+
+  // build the API string
+  const apiImgStr = JSON.stringify({
+    bucket: process.env.NEXT_PUBLIC_SERVERLESS_IMAGE_HANDLER_BUCKET,
```
namniak, line 45: as was detected by @edgardz bucket cause issues with the new CI/CD promotion approach where artifacts just move from one stage to another.
you actually don't need the bucket specified as AWS Serverless Image has it set up based on it's own env variable for each stage. So FE can manipulate it but it's optional and certainly causes issues for pre-rendered pages and artifact promotion.

@grez96 to be removed and tested

namniak (changes requested):
@edgardz you mentioned there are some issues with resizing - please provide @grez96 with details to test and address

src/components/ResponsiveImage/ResponsiveImage.tsx
grez96: Great find! I have tested removing the bucket parameter in my sandbox repository and it works perfectly! I have updated this PR with the change.

iranreyes:
@grez96 Closing this PR in favor of https://github.com/Jam3/nextjs-boilerplate/pull/269, please review if we missed something

PR #231, opened 2022-06-06
fix: replace isWindow with jam3 detect

DonghyukJacobJang:
Jira Ticket ID: <JIRA Ticket ID> 

- [ ] Code style update
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] N/A

**Aditional comments:**


src/pages/_app.tsx
```diff
@@ -13,9 +14,7 @@ import setBodyClasses from '@/utils/set-body-classes';
 
 import { store } from '@/redux';
 
-const isBrowser = typeof window !== 'undefined';
-
-if (isBrowser) {
+if (device.browser) {
```
namniak, line 17: whenever an external file is imported, Webpack won't be able to evaluate its value as false (even when it's resolved as false value in that separate file because Webpack works on a file-by-file basis).
so this will work as condition FE but Webpack won't strip out these dependencies from the prod bundle which should be seen in the bundle analyzer.

for the same reason we can't move `isBrowser = typeof window !== 'undefined'` or `isDev = process.env.NODE_ENV === 'development'` into settings as they have to be used inline or defined at the top of each file otherwise it's never false from Webpack POV at the build time


so in this case `default-passive-events` and `focus-visible` will make it to the `server` bundle.

DonghyukJacobJang: I remembered your concerns about webpack.

I don't you if you had a chance to look at `@jam3/detect` repo and how `browser` and `node` instances are initiated. As far as I see, I believe this would work because the condition is based on `OS` platfrom. [reference link](https://github.com/Jam3/detect/blob/master/src/index.ts#L170)

Because of my short experience with webpack, I was not able to test your concern. Would you mind guiding me how to test the webpack case?

src/pages/_app.tsx
```diff
@@ -13,25 +13,24 @@ import setBodyClasses from '@/utils/set-body-classes';
 
 import { store } from '@/redux';
 
-const isBrowser = typeof window !== 'undefined';
+import { isProd } from '@/data/settings';
 
-if (isBrowser) {
```
DonghyukJacobJang, line 18: Removing this condition is fine. Having them without this condition threw a `window is undefined` error a long time ago. It does not occur anymore. In addition to that, those module must be added to the production build too.

src/components/Layout/Layout.tsx
```diff
@@ -66,7 +68,7 @@ function Layout({ children }: Props) {
         />
       )}
 
-      {process.env.NEXT_PUBLIC_ENVIRONMENT !== 'production' && <AppAdmin />}
+      {!isProd && <AppAdmin />}
```
namniak, line 71: i think a lot of people prefer to show it on dev/staging otherwise admin panel is kinda useless if it's only on local so I suggest to keep NEXT_PUBLIC_ENVIRONMENT vs NODE_ENV check

DonghyukJacobJang: good find. I made a mistake on this.

alemesa (dismissed):
Looks good @DonghyukJacobJang I tested it locally and different builds, not sure about the webpack resolution tho, might be good to wait for Vadim confirmation there

src/pages/_app.tsx
```diff
@@ -13,25 +13,24 @@ import setBodyClasses from '@/utils/set-body-classes';
...
-      if (process.env.NODE_ENV !== 'production' && window.location.href.indexOf('?nostat') === -1) {
-        require('@jam3/stats')();
-      }
+    setBodyClasses();
+  }, []);
 
-      setBodyClasses();
+  /** NOTE: this is where dev tools and helper modules can be placed */
+  useEffect(() => {
+    if (!isProd && window.location.href.indexOf('?nostat') === -1) {
```
namniak, line 32: because it's not inline condition webpack will require anything that it placed here in the SSR `_app.js`
 
image

if we want to exclude this, it needs to stay with inline `process.env.NODE_ENV !== 'production'` condition rather than using `isProd` evaluated in a separate file as we proved Webpack works line-by-line in this case when stripping code

test with `npm run build:ssr` and look at `.next/server/pages/-app.js`

src/utils/why-did-you-render.ts
```diff
@@ -1,6 +1,9 @@
 import React from 'react';
+import { device } from '@jam3/detect';
 
-if (process.env.NODE_ENV === 'development' && typeof window !== 'undefined') {
+import { isDev } from '@/data/settings';
+
+if (isDev && device.browser) {
```
namniak, line 6: same thing here - this won't be excluded from SSR without inline condition
image

test with `npm run build:ssr` and look at `.next/server/pages/-app.js`

namniak: this will actually retain in `_app` client chunk to which we don't want
image

test with `npm run build:static` and npm bundle analyzer enabled
 
yu need to use import vs require - what you did for stats so that it stays in it's own chunk

namniak (changes requested):

src/utils/why-did-you-render.ts
DonghyukJacobJang: I tried to lazy load the `why-did-you-render` library and it failed due to the way of its mechanism. To secure and clean our production bundle, I am going to put the condition back which is `process.env.NODE_ENV === 'development'`.

namniak (approved):

PR #260, opened 2022-08-30
[BUG] Cache Invalidation for Git Flow Workflow

grez96:
Noticed this in my project (using git flow). Cache wasn't invalidating and taking a long time for changes to reflect on environment after successful deploy. Cloudfront had no invalidation. Simply copy and pasted the invalidation step from trunk based workflow to the git flow one.


iranreyes:
Not relevant, will be improved

PR #263, opened 2022-09-16
feature: add owasp zap as dast tool

iranreyes:
A working implementation of DAST - Owasp ZAP. It scans the staging environment with a full scan by default.

The report will be uploaded to the development S3 bucket inside `/report/security/dast`. We will keep all the reports in a hierarchy inside `/report`.

Added optional ZAP config files for the rules and work-in-progress tickets.

> Also added basic auth token variables depending on the environment, as they are different.


alemesa:
Closing this since the PR has been open for more than 1 year
Let's re-open if this gains traction at some point
Thx

PR #264, opened 2022-09-19
feat: add html validator

iranreyes:
A working implementation of HTML 5 Validator. It scans the built files (pre-rendered) and generates a raw report

The report will be uploaded to the development S3 bucket inside /report/html. We will keep all the reports in a hierarchy inside /report.


alemesa:
Closing this since the PR has been open for more than 1 year
Let's re-open if this gains traction at some point
Thx

PR #276, opened 2023-02-21
Update use-cookie-banner.ts

namniak:
Fixing re-render issue showing on route change by using `useCallback` to make sure functions are persistent 
image


alemesa (approved):
🙏

PR #278, opened 2023-02-24
Storybook update

namniak:
* **update Preview:**  decorators and parameters to match Nextjs app setup
* **update Webpack:** We can make it simple and eliminate the SVG path and only care about the extension.
This will allow injecting SVG as code from any folder and not stick to `components/svgs` this can be handy esp when bringing things from React UI library where icons can be contained within a component folder e.g VideoPlayer 
This is how our next webpack config is set up as well without specifying `components/svgs` path and simply looking at the SVG extension


alemesa (approved):
🙏

PR #290, opened 2023-05-18
PATCH-1: Update packages

edgardz:
- Reinstall all NPM packages from scratch (to get rid of all the old packages)
- Migrate impacted code to work with the newer versions (including React 18 and Next 13)
- Remove all unused packages and packages we always remove anyways
- Remove @jam3 packages (they are outdated and will be reintroduced later on with a monorepo approach)
- Update linter rules, they were really outdated!. In the process also removed the old and unnecessary semi-colons.
- Enable Git LFS by default
- Storybook V7


src/components/AppAdmin/AppAdmin.tsx
gm-santos, line 1: a suggestion, a button that we could remove for once the app admin, using some locaStorage/cookie?

edgardz: Yup. This is coming on next patches. This PR is to prepare the ground for upcoming updates.

gm-santos (dismissed):
looking good IMO, for those who will review this, 90% of the changes are removing the `;`

gm-santos (approved):

PR #291, opened 2023-05-19
PATCH-2: View/Controller design pattern

edgardz:
Builds on top of #290

- Update AppAdmin
- Use CircleCI pipeline number as build number
- Runtime environment detection (since we are using trunk based development, we cannot rely on env vars to detect the environment)
- Move Svgs to separate folder
- Refactor page components to use the View/Controller design pattern
- Update generator to reflect the new pattern
- Page transitions
- Refactor content injection (CMS groundwork)
- Optimize images based on component size
- Calculate static image sizes
- Update Readme


edgardz:
TIP: To bypass the preview credentials, use this link https://test%20env:password@preview-bae8179.dev.nextjs-boilerplate.jam3.cloud.jam3.net/

src/data/config.json
```diff
@@ -0,0 +1,22 @@
+{
```
gm-santos, line 1: this is awesome

alemesa (dismissed):
amazing!

🧑‍🍳😙🤌

src/components/Layout/Layout.tsx
```diff
@@ -44,23 +52,45 @@ const Layout: FC<AppProps<PageProps>> = ({ Component, pageProps }) => {
     }
   }, [router.events, handleRouteChange])
 
+  // handle page transitions
   useEffect(() => {
-    checkWebpSupport('lossy', (isSupported) => dispatch(setIsWebpSupported(isSupported)))
-  }, [dispatch])
+    const transitionTimeline = gsap.timeline()
```
gm-santos, line 57: 🤯

gm-santos (dismissed):
transition in/out when navigate 🤯

src/components/Head/Head.tsx
```diff
@@ -1,64 +1,70 @@
...
         <>
-          <link rel="preconnect" href={process.env.NEXT_PUBLIC_DNS_PREFETCH} crossOrigin="anonymous" />
-          <link rel="dns-prefetch" href={process.env.NEXT_PUBLIC_DNS_PREFETCH} />
+          <link key={href} rel="preconnect" href={href} crossOrigin="anonymous" />
+          <link rel="dns-prefetch" href={href} />
         </>
-      )}
+      ))}
 
       {process.env.NODE_ENV === 'development' && (
```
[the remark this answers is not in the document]
edgardz: The Head component is a bit different. Everything here must be rendered server side so the crawlers have the correct information when they load the HTML. Since we are using a trunk approach now, it means we have a single build for all envs, so the most we can do is detect if we are running the application on dev or production mode. We cannot detect the runtime environment at build time.

src/components/AppAdmin/AppAdmin.tsx
```diff
@@ -1,76 +1,130 @@
...
+export interface ViewProps extends AppAdminProps {}
+
+// View (pure and testable component, receives props from the controller)
+export const View: FC<ViewProps> = ({ className }) => {
+  const { width, height } = useWindowSize()
+
+  const env = useMemo(() => getRuntimeEnv(), [])
+  const date = useMemo(() => process.env.NEXT_PUBLIC_COMMIT_DATE || '', [])
+  const commit = useMemo(() => process.env.NEXT_PUBLIC_COMMIT_ID || '', [])
+  const version = useMemo(() => process.env.NEXT_PUBLIC_VERSION_NUMBER || '', [])
```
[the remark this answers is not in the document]
edgardz: Actually. We should! 😅

alemesa:
looking great, I cloned it and played around a bit

found only one more thing @edgardz, the cookie banner is not clickable

maybe we can just add in the `$elements` array => `cookie` after `pages`
and then just `@include z-index(cookie)`

alemesa (approved):
✨

.storybook/intro/Svg.stories.tsx
```diff
@@ -30,7 +30,7 @@ const SvgCatalog: FC<SvgCatalogProps> = ({ viewbox, color }) => {
       {icons.map((icon, i) => {
         return (
           <div
-            key={i}
+            key={icon.name}
```
alemesa, line 33: sonar is pretty useless sometimes 💀

namniak (changes requested):
@edgardz we need to change back the assets folder and make sure `public/assets` and  `public/favicons` are bac under `public/common`.
This was done for a reason as we have CloudFront `/common` behavior with a specific caching policy attached

src/components/PageHome/PageHome.module.scss
```diff
@@ -1,6 +1,10 @@
 @import 'shared';
 
 .root {
+  position: relative;
+  opacity: 0;
```
namniak, line 5: same

src/components/PageNotFound/PageNotFound.module.scss
```diff
@@ -1,15 +1,12 @@
 @import 'shared';
 
-// IMPORTANT: use safe styles here. avoid flex and other latest CSS as this page is intended to be shown in legacy browsers aka IE9/10/11
-
 .root {
-  text-align: center;
+  position: relative;
+  opacity: 0;
```
namniak, line 5: same

src/components/PageUnsupported/PageUnsupported.module.scss
```diff
@@ -3,13 +3,12 @@
 // IMPORTANT: use safe styles here. avoid flex and other latest CSS as this page is intended to be shown in legacy browsers aka IE9/10/11
 
 .root {
-  text-align: center;
+  position: relative;
+  opacity: 0;
```
namniak, line 7: same

.circleci/scripts/deploy.sh
```diff
@@ -10,12 +10,12 @@ if [ "$CI_ENV" == "development" ]; then
 fi
 
 # Sync bundles with strong cache
-aws s3 sync ./common/favicons s3://${S3_ORIGIN_BUCKET}/common/favicons --metadata-directive 'REPLACE' --cache-control max-age=31536000,public ${DELETE_OLD_FILES}
+aws s3 sync ./favicons s3://${S3_ORIGIN_BUCKET}/favicons --metadata-directive 'REPLACE' --cache-control max-age=31536000,public ${DELETE_OLD_FILES}
```
namniak, line 13: revert paths in this file

.storybook/manager-head.html
```diff
@@ -1 +1 @@
-<link rel="shortcut icon" href="common/favicons/favicon.ico" />
+<link rel="shortcut icon" href="favicons/favicon.ico" />
```
namniak, line 1: pls revert

.storybook/manager.ts
```diff
@@ -6,6 +6,6 @@ addons.setConfig({
     base: 'dark',
     brandTitle: 'Jam3 NextJS Boilerplate',
     brandUrl: 'https://jam3.com',
-    brandImage: '/common/favicons/favicon-32x32.png'
+    brandImage: '/favicons/favicon-32x32.png'
```
namniak, line 9: pls revert

public/favicons/site.webmanifest
```diff
@@ -3,17 +3,17 @@
   "short_name": "React App",
   "icons": [
     {
-      "src": "/common/favicons/android-chrome-192x192.png",
+      "src": "/favicons/favicon-192x192.png",
```
namniak, line 6: pls revert this file

src/pages/_app.tsx
```diff
@@ -21,17 +21,30 @@ gsapInit()
 // This default export is required in a new `pages/_app.js` file.
 const App: FC<AppProps<PageProps>> = (props) => {
   useEffect(() => {
+    // Body class names
     setBodyClasses()
+
+    // Disable automatic scroll restoration
+    window.history.scrollRestoration = 'manual'
```
namniak, line 28: why do we wanna do that?
this will prevent page restoration when you use the browser's back button which is pretty bad for the cases when you are on the blog page and click one of the blog cards then use browser back button and expect to end up where you were in the grid but you end up at the hero instead

src/pages/_app.tsx
```diff
@@ -21,17 +21,30 @@ gsapInit()
...
+      node.removeAttribute('data-n-g')
+      node.removeAttribute('data-n-p')
+    })
+
+    new MutationObserver((mutations) => {
+      mutations.forEach(({ target }) => {
+        const t = target as Element
+        if (t.nodeName === 'STYLE' && t.getAttribute('media') === 'x') t.removeAttribute('media')
+      })
+    }).observe(document.head, { subtree: true, attributeFilter: ['media'] })
```
namniak, line 41: We initially saw this issue back in 2020 and we had this fix added to one of the projects. is it still an issue? Haven't seen it in a while. I guess it might still happen in the case when the route changes and you still have the old page on the screen animating out in a transition group - we just probably didn't have this case for a long time

src/data/config.json
```diff
@@ -0,0 +1,16 @@
...
+  "siteDescription": "Default Description",
+  "siteKeywords": ["Jam3", "Web App", "React", "NextJS"],
+  "websiteUrl": "https://localhost:3000",
+  "dnsPrefetch": {
+    "development": [],
+    "production": ["https://www.google-analytics.com", "https://www.googletagmanager.com"],
+    "test": []
+  },
+  "analytics": {
+    "gtmIds": { "prod": "GTM-0000000", "stage": "GTM-0000000", "dev": "GTM-0000000", "local": "" }
```
namniak, line 14: i suggest also adding GTM params per env as we recently often have one GTM property where ID is the same but params a re different based on the environments

src/utils/analytics.tsx
```diff
@@ -56,7 +68,7 @@ export function GtmScript({ consent }: Props) {
         new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
         j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
         'https://www.googletagmanager.com/gtm.js?id='+i+dl${QUERY_PARAMS};f.parentNode.insertBefore(j,f);
-        })(window,document,'script','dataLayer', '${process.env.NEXT_PUBLIC_GTM_ID}');
+        })(window,document,'script','dataLayer', '${gtmId}');
```
namniak, line 71: add params to the URL is case it's needed

src/pages/_document.tsx
```diff
@@ -19,9 +10,7 @@ class MyDocument extends Document {
     return (
       <Html lang="en">
         <Head />
-        {hideStaticHtml && (
-          <script data-cfasync="false" dangerouslySetInnerHTML={{ __html: sanitizer(hideStaticHtmlScript) }} />
-        )}
+
```
namniak, line 13: pls add the code back it's a much better approach than adding opacity 0 to page and navs. this is actually doing what it's supposed to do - have all content visible on the server (e.g. with disabled JS) and removes that flash if FE is detected as the call is added to the document so the whole thing is hidden until the app kicks in and removes this class

src/utils/set-body-classes.ts
```diff
@@ -1,4 +1,4 @@
-import { hideStaticHtml } from '@/data/settings'
+import ResizeService from '@/services/resize'
```
namniak, line 1: please add the static HTML back. it's disabled by default if you don't want it but I actually recommend enabling that by default rather than removing

with the change made in this PR you's always see a blank page with JS disabled when buit and served:
image

namniak (changes requested):

edgardz:
@namniak Oh! I was not aware there was a logic coupling with code from another repo!

src/components/Nav/Nav.module.scss
```diff
@@ -11,6 +11,7 @@
   box-shadow: $element-shadow;
   background-color: $white;
   height: $navbar-height-mobile;
+  opacity: 0;
```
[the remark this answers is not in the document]
edgardz: Actually, when JS is disabled a proper <noscript> screen should be rendered.

src/components/PageAbout/PageAbout.module.scss
```diff
@@ -1,10 +1,12 @@
 @import 'shared';
 
 .root {
+  position: relative;
+  opacity: 0;
```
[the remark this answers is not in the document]
edgardz: The past solution was still prone to FOUC on slow connections since it requires the JS to load and be executed in order to hide content. The new approach brings all pages hidden by default and unhide them when the JS kicks in. For the case of JS disabled browsers, we render a proper <noscript> screen.

src/pages/_app.tsx
edgardz: This prevents the page from scrolling before the transition out ends. If we don't have this, the scroll would jump before the current page have transitioned out.

src/pages/_app.tsx
edgardz: Yup. This is here because of page transitions.

src/utils/analytics.tsx
edgardz: Refactored this util to be a service.

src/pages/_document.tsx
edgardz: This solution is prone to FOUC on slow connections as mentioned before.

src/utils/set-body-classes.ts
edgardz: Now you see this:
image

src/pages/_document.tsx
namniak: keep in mind`hideStaticHtml` is set to `false` out of the box so there's def hydration visible but with that on it's better

src/utils/set-body-classes.ts
namniak: yes let's discuss this in the future
@iranreyes @DonghyukJacobJang @edgardz we need to decide what the future most common scenario is and just agree on the approach
both approaches have minor issues but for majority of the demographic either way works and NO-JS splash screen is less work compared to making static pages look good for every single component on it and with relatively small number of crawlers not executing JS maybe that's ok to sacrifice that? I'm 50/50 on this when thinking of different types of apps

src/data/config.json
namniak: lovely!

src/utils/set-body-classes.ts
edgardz: Added `supportsNoJs` flag to data/config.json. This will allow us to chose the behavior depending on the project. if it is set to true, pages will not be hidden by default.

src/pages/_document.tsx
```diff
@@ -19,8 +14,20 @@ class MyDocument extends Document {
     return (
       <Html lang="en">
         <Head />
-        {hideStaticHtml && (
-          <script data-cfasync="false" dangerouslySetInnerHTML={{ __html: sanitizer(hideStaticHtmlScript) }} />
+        {config.supportsNoJs && (
```
namniak, line 17: @edgardz I don't think it needs to be connected to `supportsNoJs`.
the idea of `hideStaticHtml` was that with JS available HTML is hidden until the app kicks in and unhides it so there's no static flash while  `supportsNoJs` needs to only control `ScreenNoScript` functionality.

Injection of these particular styles still needs to be based on `typeof window !== 'undefined'` condition and apply regardless of `supportsNoJs` value.
idk if we wanna always do that by default (ideally yes) but If we don't then we need to re-introduce a separate `hideStaticHtml` option.
if we introduce a separate flag then this only needs to apply based on the flag in Layout.js
```
  // make sure the page us un-hidden once application kicks in
  useEffect(() => {
    gsap.set(document.documentElement, { autoAlpha: 1 })
    gsap.set(rootRef.current, { autoAlpha: 1 })
  }, [])
```

I know that we now hide the pages with JS on mount via hook but the difference is that those would be only executed once the whole app kicks in (meaning there will be a flash until then) while the hide should apply asap which is what the script injection will take care of.

src/components/Layout/Layout.tsx
```diff
@@ -44,34 +63,99 @@ const Layout: FC<AppProps<PageProps>> = ({ Component, pageProps }) => {
...
   useEffect(() => {
-    checkWebpSupport('lossy', (isSupported) => dispatch(setIsWebpSupported(isSupported)))
-  }, [dispatch])
-
-  return (
-    <>
-      <GtmScript consent={cookieConsent?.statistics} />
-
-      <Head {...pageProps.head} />
+    gsap.set(document.documentElement, { autoAlpha: 1 })
```
namniak, line 68: See comment about `hideStaticHtml` condition for this.

src/components/Layout/Layout.module.scss
```diff
@@ -0,0 +1,17 @@
+@import 'shared';
+
+.root {
+  display: flex;
+  flex-direction: column;
+
+  @if $supports-no-js != true {
+    visibility: hidden;
+    opacity: 0;
+  }
```
namniak, line 10: based on testing this doesn't seem to apply when `supportsNoJs` is set to `true`.
Also not sure it's needed if we already apply hide to the whole HTML via script

src/components/Layout/Layout.tsx
```diff
@@ -44,34 +63,99 @@ const Layout: FC<AppProps<PageProps>> = ({ Component, pageProps }) => {
...
+    gsap.set(rootRef.current, { autoAlpha: 1 })
```
namniak, line 69: probably extra see comment above

src/components/Layout/Layout.tsx
```diff
@@ -1,30 +1,49 @@
...
+
+  const rootRef = useRef<HTMLDivElement>(null)
+  const navHandleRef = useRef<NavHandle | null>(null)
+  const pageHandleRef = useRef<PageHandle | null>(null)
+  const isFirstPageRef = useRef(true)
+  const isGoingBackRef = useRef(false)
+  const scrollHistoryRef = useRef<{ pathname: string; value: number }[]>([])
+  const currentPathnameRef = useRef(router.asPath.split('#')[0].split('?')[0])
+  const scrollRestorationTimeoutRef = useRef<NodeJS.Timeout>()
+
```
namniak, line 46: this works well but there's a catch it stores one page at a time and loses the position once you scroll on the new page. e.g. 

- land on about scroll, go to index, use the back button - about is restored as expected
- land on about scroll, go to index and scroll, use the back button - about page position is lost.

Native page restoration remembers the position per page and switching between two pages using the browser's back button on each restore position independently.

It might be worth extending Redux functionality (which already stores `prevRoute`) and store page positions on leave and then applying it here on the Layout level on route change if it's `going back`

namniak (changes requested):
Tested new changes and we are almost there!

namniak (approved):
Thanks, @edgardz!
A lot of great stuff here - appreciate your work.
Let's 🚢 it

@DonghyukJacobJang let us know if you wanna have a look first.
I think this is important to get this in asap

alemesa (approved):
🔥 🚀

PR #297, opened 2023-08-31
feat: BaseCarousel

alemesa:
* BaseCarousel 

* **Usage**
```
<BaseCarousel>
    <div>Slide 1</div>
    <div>Slide 2</div>
    <div>Slide 3</div>
</BaseCarousel>
```

* Maybe it will be worth to implement Navigation, Pagination, ProgressBar, ScrollBar but thinking it might be bloated, usually we have simple carousel with some arrows (open to opinions)

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "react-dom": "18.2.0",
     "react-redux": "^8.0.5",
     "string-format": "^2.0.0",
+    "swiper": "^10.1.0",
     "xss": "^1.0.14"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -63,6 +63,7 @@
     "react-dom": "18.2.0",
     "react-redux": "^8.0.5",
     "string-format": "^2.0.0",
+    "swiper": "^10.1.0",
```
edgardz, line 66: Is this Swiper version working well on mobile? I remember I had to downgrade swiper cause they removed touch events in favour of pointer events. At the time, pointer events didn't work well on some mobile devices.

alemesa: I'll test this @edgardz didn't know 👍

fzhang13:
This is more of a thought, but should we be moving component into the [next-react-ui](https://github.com/Jam3/next-react-ui) repo? I feel like component wise we're getting too bloated. Even though I know we use a lot of Carousels on projects sadly

PR #298, opened 2023-09-15
[Feature] Add AWS RUM Integration

alex-togo:
This PR adds AWS RUM (real user monitoring) to the frontend boilerplate.

Env vars necessary to be added to CircleCI are found in the JS snippet given to you on the RUM page in the AWS management console. Since there are many env vars, we opted to use a single env var with JSON encoded values. 

Here is an example of how the env var would look:
```
NEXT_PUBLIC_AWS_RUM='{"guestRoleArn":"arn:aws:iam::12345:role/cognito_unauthenticated_role","identityPoolId":"us-east-1:xxxxxx-xxxxx-xxxxx","endpoint":"https://dataplane.rum.us-east-1.amazonaws.com","region":"us-east-1","appId":"xxxxxxx-xxxx-xxxx-xxxx-xxxxxx","appVersion":"1.0.0"}'
```

This code will also record the user's journey, based on their navigation through the app using Next.js router.

NPM Page: https://www.npmjs.com/package/aws-rum-web
Bundlephobia minified size: https://bundlephobia.com/package/aws-rum-web@1.15.0

User Journey Example (Home -> About -> Home):
image

**Note:** If the RUM service is not setup to run on localhost, you will get a 400 error. You'll need to use the devops boilerplate to set the domain as localhost. Otherwise, it will run on your specified domain.


package.json
```diff
@@ -48,6 +48,7 @@
   ],
   "dependencies": {
     "@reduxjs/toolkit": "^1.9.5",
+    "aws-rum-web": "^1.14.0",
```
edgardz, line 51: Hey @alex-togo can you check this package size? The NPM page says 1.27 MB (!!!) Is this true?

fzhang13: minified version seems to be 136kb according to bundlephobia

alex-togo: @edgardz yeah the unpacked size is currently 1.31mb according to the NPM site, and 136.2kb minifed as Felix mentioned. Is the unpacked size going to be a problem with this package?

edgardz (changes requested):

.env.production
```diff
@@ -1,5 +1,16 @@
 OUTPUT=export
 BUNDLE_ANALYZE=false
+NEXT_PUBLIC_ALLOW_AWS_RUM=true
+NEXT_PUBLIC_GUEST_ROLE_ARN=$RUM_GUEST_ROLE_ARN
+NEXT_PUBLIC_IDENTITY_POOL_ID=$RUM_IDENTITY_POOL_ID
+NEXT_PUBLIC_ENDPOINT=$RUM_ENDPOINT
+NEXT_PUBLIC_APP_ID=$RUM_APP_ID
+NEXT_PUBLIC_APP_VERSION=$RUM_APP_VERSION
+NEXT_PUBLIC_APP_REGION=$RUM_APP_REGION
```
edgardz, line 9: I'm considering if we should set a single env var with JSON encoded values, instead of introducing 7 new env vars. What do you think?

src/services/aws-rum.ts
```diff
@@ -0,0 +1,44 @@
...
+        identityPoolId: process.env.NEXT_PUBLIC_IDENTITY_POOL_ID,
+        endpoint: process.env.NEXT_PUBLIC_ENDPOINT,
+        telemetries: ['http', 'errors', 'performance'],
+        allowCookies: cookieConsent,
+        enableXRay: false,
+        disableAutoPageView: true
+      }
+
+      const APPLICATION_ID = process.env.NEXT_PUBLIC_APP_ID || ''
+      const APPLICATION_VERSION = process.env.NEXT_PUBLIC_APP_VERSION || ''
```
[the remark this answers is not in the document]
alex-togo: Yes, this is the version number of RUM.

.env.production
alex-togo: Great suggestion, I didn't even know this was possible, really cool!

src/pages/_app.tsx
```diff
@@ -40,7 +50,17 @@ const App: FC<AppProps<PageProps>> = (props) => {
         if (t.nodeName === 'STYLE' && t.getAttribute('media') === 'x') t.removeAttribute('media')
       })
     }).observe(document.head, { subtree: true, attributeFilter: ['media'] })
-  }, [])
+  }, [cookieConsent])
```
edgardz, line 53: Don't add this dependency here. The code in this effect is supposed to run only once. Create another effect instead.

edgardz (changes requested):

edgardz (approved):

PR #301, opened 2023-10-03
Remove boilerplate images from LFS

edgardz:


edgardz:
I don't know why the images don't appear in the PR diff.
They exist in the branch and in the preview, though:
https://github.com/Jam3/nextjs-boilerplate/blob/fix-lfs/docs/logo.jpg

PR #314, opened 2023-11-16
PATCH-8: Refactor view/controller + CmsService [merge after #313]

edgardz:


alemesa (approved):

alemesa:
I think it's ok for now @edgardz but maybe we can find another name for CmsService, I think it may confuse people, PageCopyService or just CopyService not sure

